PR #17, opened 2020-05-23
Initial draft of wire format messages

chris-wood:
This change shuffles around content and replaces the text-based protocol descriptions with protocol flows and messages. It also includes a first cut at wire format details for these messages. (We can always iterate later on!) 

cc @hugokraw

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -116,7 +116,6 @@ key exchange resilient to server compromise"
 
   I-D.ietf-tls-exported-authenticator:
   I-D.barnes-tls-pake:
-  I-D.irtf-cfrg-argon2:
   I-D.irtf-cfrg-hash-to-curve:
   I-D.irtf-cfrg-voprf:
   I-D.sullivan-tls-opaque:
@@ -214,7 +213,6 @@ online)"
 
   RFC2945:
   RFC5869:
-  RFC7914:
   RFC8018:
   RFC8125:
   RFC8446:
@@ -412,7 +410,7 @@ operations, roles, and behaviors of OPAQUE:
 
 - Client (U): Entity which has knowledge of a password and wishes to authenticate.
 - Server (S): Entity which authenticates clients using passwords.
-- (skX, pkX): A group key pair used in role X; skX is the private key and pkX is
+- (skX, pkX): A key pair used in role X; skX is the private key and pkX is
 the public key. For example, (skU, pkU) refers to the U's private and public key.
 - pk(skX): The public key corresponding to private key skX.
 - encode_big_endian(x, n): An byte string encoding the integer value x as an n-byte
@@ -424,9 +422,9 @@ big-endian value.
 
 Except if said otherwise, random choices in this specification refer to
 drawing with uniform distribution from a given set (i.e., "random" is short
-for "uniformly random"). All sources of randomness SHOULD adhere to the requirements
-in {{!RFC4086}}. Where noted, random choices can be replaced with fresh outputs
-from a cryptographically strong pseudorandom generator or pseudorandom function.
+for "uniformly random"). Random choices can be replaced with fresh outputs from
+a cryptographically strong pseudorandom generator, according to the requirements
+in {{!RFC4086}}, or pseudorandom function.
 
 The name OPAQUE: A homonym of O-PAKE where O is for Oblivious
 (the name OPAKE was taken).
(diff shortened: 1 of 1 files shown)


hugokraw:
Hi Chris, I went over the specification text. Looks good (and a lot of
work).
Thanks!

I have several comments. I was not sure how to better communicate this in
the
github so pardon the use of 20th century technology here...

Comments are mostly in the order of appearance in the document, not by
order of
importance.

-- You say:
> Client (U): Entity which has knowledge of a password and wishes to
authenticate.
Here and in other places, including the naming of the online phase as the
"authentication" phase, make this seem as just authentication when it is
also
key exchange. I have this note in the draft saying that authentication
without
key exchange is "analogous to carefully checking a visitor's credential at
the door and then leaving the door open for others to enter freely."

-- The expression "group key pair" was unclear/confusing to me.

-- Somewhere you say: "Where noted, random choices can be replaced with
fresh
outputs from a cryptographically strong pseudorandom generator or
pseudorandom
function."

The default is that we can replace randomness with pseudorandomness; we
should
note the cases where this is not the case and not the cases where
replacement is
fine (essentially all cases).

-- VOPRF: We never use the verifiability property; it would violate the
PKI-freeness. We can use g^kU for fixed-base blinding but g^kU in this case
is
not certified, just sent from server to client (possibly cached by the
latter)

-- Fixed-base blinding: You are defining the OPRF with var-base blinding,
not
leaving the option of fixed-base blinding (which would need to transfer
g^kU).

-- Generic OPRF vs DH-OPRF: This issue of blinding mechanisms points to the
fact
that while all components of the protocol are generic in your
specification, the
OPRF is fixed to DH-OPRF. I do not suggest complicating presentation too
much
just for the sake of being generic but I would definitely want to allow an
optimization such as fixed-base blinding, and this may call for a more
general
API (e.g., an OPRF structure with Request/Response).

-- From the expression Open(k, n, aad, ct) I understand that you do not see
aad
as part of ct. Is this right?

-- In the OPRF ciphersuites there is no 256-bit curve. Why is that?
In the case of integration with TLS wouldn't we be using the curves TLS
uses?

-- In the hardening functions, shouldn't one allow for PBKDF for backward
compatibility reasons? I know of some company that wanted that exactly for
that
reason.

-- The Credentials structure: Setting it to the 4 elements as you define is
fine
and should cover common cases. However, pkU is usually not needed by the
client
(that's the case for all the AKE examples I considered) so it may be
removed.
The public key pkS need not be encrypted but I assume we don't gain much by
not
encrypting it. Do you mean to always encrypt all elements? And if so, why is
there a AAD field?

I know we do not want to leave too many options but I would not want to
prohibit
the encryption-less case where you only send the authenticated pkS (skS is
derived from RwdU). The only place where I saw some benefit in AuCPace is
that
they had less bits in the wire. OPAQUE with the encryption-less option beats
them in that sense too. This mode also removes the annoying equivocable
encryption property that we need in the UC analysis.

--  When specifying the envelope encryption, I defined an optional key KdKey
derived from RwdU for extended functionalities, e.g., as a way to retrieve
non-OPAQUE credentials (or to support the encryption-less option). We will
see
how we go about specifying the envelope encryption and then see what
happens to
such KdKey.

-- RwdU = Harden(y): This function should accept hardness parameters (or you
assume they are part of the function definition)

-- The name OPAQUE collides with "opaque" as an object type. Maybe say at
the
beginning that the name OPAQUE is always capitalized while the object type
is
not (or don't say anything and trust the intelligence of the reader :-)

-- You ask: should we blind both the identity and password?
   Is the question whether IdU should be part of the input to the OPRF
(together
   with PwdU)? It is not necessary, I think, because kU is supposed to be
unique
   per user. Also, servers may use more than on IdU value to identify a
user or
   use a value of IdU different than the one the user has (but we need some
   agreement on the value of IdU for the sake of key derivation so maybe
this
   would not be a problem if we wanted to use IdU as input to the OPRF).

-- At registration the use chooses fresh keys (skU, pkU) =
GenerateKeyPair().
   Would it be possible that the user has a specific pair it wants to
register?
   Maybe because it has a hardware token with these keys? But still uses the
   envelope in case he left them at home...? WebAuth related?

-- You say (regarding IdS): "we should probably improve this a bit and
provide
   guidance for where and how IdS is provisioned to clients"
   Probably such guidance is needed although there may be many use cases.
One
   place where this id can be communicated is in the Response message from
the
   server or in the envelope.

-- Regarding the second stage you say: "This stage is composed of a
sequential
OPRF and key exchange flow."
Do not use the word sequential as we run them in parallel. In integration
with
TLS we even start the KE part before the OPRF (exchanging DH values).

-- "client and server agree on the knowledge of the password and have
mutually
authenticated one another"
This is more like: Client proves user's knowledge of the password and agree
on a
mutually authenticated shared secret key.

-- In the online stage, should IdS be (at least an optional) input to the
client? It needs it to contact the server. Otoh, the name IdS used later in
the
protocol may be different than the one the client initially has for
contacting
the server, especially if IdS is communicated in the envelope.

-- the context value is always derived as fresh randomness. However, it
would
make sense to set this value to the KE nonces (particularly in the case of
TLS).

-- Specification of the key-exchange protocol is out of scope for this
document.
There is one point where the two components, OPRF and AKE, need to be bound
together. This can be done by adding OPRF1 (what I call the value alpha) to
the
data authenticated by the KE. In the current specific, as I wrote it, all
of the
OPRF transcript (and even the envelope) is authenticated, not because this
is
needed but because it is simpler to just add everything to the stream of
authenticated data. Anyway, the point is that this binding needs to be
accommodated in gluing the OPRF and KE.

-- Generic committed encryption. I am not convinced that the use of RKR here
justifies a separate document. The main reason for having such document is
because the trivial solution, namely, encrypt-then-CRmac (CRmac is any MAC
that
is collision resistant), is slow for large pieces of data which is not the
case
here. I am afraid that such separate document will have much more complexity
that what's needed here, and we will end recommending using
encrypt-then-hmac
for OPAQUE (it is the most robust mode for RKR encryption). I would have
hoped
to have a spec for it already but even defining it specifically for OPAQUE
is
not a big deal. One issue is that given the popularity of GCM, people will
want
to use it as the encryption part.  This is not a problem security-wise but
wastes computation and bandwidth (not sure how serious this is). That's why
I
defined standalone counter mode encryption (another self-contained spec I
could
not find).

PR #33, opened 2020-06-04
Fix and relocate KdKey derivation.

chris-wood:
This moves derivation of the KdKey to the main protocol flow (outside of the RKR-AEAD algorithm), uses HKDF-Extract for secret derivation, and renames KdKey to exporter_key.


chris-wood:
> I want to beef up the explanation of how the exporter_key may be used but no need to keep this open now.

Yeah, more text would certainly be useful. I figured we could do that in another PR. :-)

PR #36, opened 2020-06-17
Remove RKR-AEAD dependency.

chris-wood:
This removes the RKR-AEAD dependency in favor of a simpler OTP-like construction for credential secrecy. It also replaces redundant Extract calls on RwdU with Expand calls (RwdU is already assumed to be the output of a RO). Finally, it removes the `context` parameter from all protocol messages, punting that functionality to applications.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -571,7 +571,7 @@ multiple users. These steps can happen offline, i.e., before the registration ph
 Once complete, the registration process proceeds as follows:
 
 ~~~
- Client (IdU, PwdU, skU, pkU)                 Server (skS, pkS)
+ Client (IdU, PwdU, aad, skU, pkU)                 Server (skS, pkS)
   -----------------------------------------------------------------
    request, metadata = CreateRegistrationRequest(IdU, PwdU)
 
@@ -583,7 +583,7 @@ Once complete, the registration process proceeds as follows:
                                    response
                               <-----------------
 
-    record = FinalizeRequest(IdU, PwdU, metadata, request, response)
+ record = FinalizeRequest(IdU, PwdU, aad, skU, metadata, request, response)
 
                                     record
                               ------------------>
@@ -675,7 +675,7 @@ CreateRegistrationResponse(request, pkS)
 
 Input:
 - request, a RegistrationRequest structure
-- pkS, the server's public KE key
+- pkS, the server's public key
 
 Output:
 - response, a RegistrationResponse structure
@@ -693,7 +693,7 @@ Steps:
 #### FinalizeRequest
 
 ~~~
-FinalizeRequest(IdU, PwdU, skU, metadata, request, response)
+FinalizeRequest(IdU, PwdU, aad, skU, metadata, request, response)
 
 Parameters:
 - params, the MHF parameters established out of band
@@ -703,6 +703,8 @@ Parameters:
 Input:
 - IdU, an opaque byte string containing the user's identity
 - PwdU, an opaque byte string containing the user's password
+- aad, application-specific additional associated data
+- skU, the user's private key
 - metadata, a RequestMetadata structure
 - request, a RegistrationRequest structure
 - response, a RegistrationResponse structure
@@ -723,24 +725,24 @@ Steps:
 9. auth_key = HKDF-Expand(RwdU, contact(nonce, "AuthKey"), Nk)
 10. exporter_key = HKDF-Expand(RwdU, concat(nonce, "ExporterKey"), Nk)
 11. ct = xor(pt, pseudorandom_pad)
-12. t = HMAC(auth_key, concat(nonce, ct, concat(pkS, aad))), where aad is
-application-specific additional associated data.
-13. EnvU = concat(nonce, ct, aad, t)
-14. Create RegistrationUpload upload with envelope value (EnvU, pkU).
-15. Output (upload, exporter_key)
+12. auth_data = pkS
+13. t = HMAC(auth_key, concat(nonce, ct, concat(auth_data, aad)))
+14. EnvU = concat(nonce, ct, aad, t)
+15. Create RegistrationUpload upload with envelope value (EnvU, pkU).
+16. Output (upload, exporter_key)
 ~~~
 
 [[RFC editor: please change "RFCXXXX" to the correct number before publication.]]
 
 Applications MUST encrypt and authenticate skU, and MUST authenticate pkS.
 Secrecy of pkS is optional. If an application requires secrecy of pkS, this value
-SHOULD be included in the Credentials structure (step 5) and omitted from the HMAC
-computation (step 12).
+SHOULD be included in the Credentials structure (step 5) and omitted from
+auth_data (step 12).
 
 Applications may optionally include pkU, IdU, or IdS in the Credentials structure
-at step 5 if secrecy of these values is desired. If an application does not encrypt any
-of these values, they may be included in the aad. Instantiations of OPAQUE MUST specify
-how aad is constructed and serialized.
+(step 5) if secrecy of these values is desired. If an application does not require secrecy
+for these values, they may be appended to auth_data (step 12). Instantiations of OPAQUE
+MUST specify how auth_data is constructed.
 
 The server identity `IdS` comes from context. For example, if registering with
 a server within the context of a TLS connection, the identity might be the
@@ -778,7 +780,7 @@ used in this stage.
                                    response
                               <-----------------
 
-    creds = RecoverCredentials(PwdU, metadata, request, response)
+    creds = RecoverCredentials(PwdU, aad, metadata, request, response)
 
                                (AKE with creds)
                               <================>
@@ -880,10 +882,10 @@ Steps:
 Applications which include pkS in the Credentials structure MAY omit
 pkS from the CredentialResponse.
 
-#### RecoverCredentials(IdU, PwdU, metadata, request, response)
+#### RecoverCredentials(PwdU, aad, metadata, request, response)
 
 ~~~
-RecoverCredentials(PwdU, metadata, request, response)
+RecoverCredentials(PwdU, aad, metadata, request, response)
 
 Parameters:
 - params, the MHF parameters established out of band
@@ -892,6 +894,7 @@ Parameters:
 
 Input:
 - PwdU, an opaque byte string containing the user's password
+- aad, application-specific additional associated data
 - metadata, a RequestMetadata structure
 - request, a RegistrationRequest structure
 - response, a RegistrationResponse structure
@@ -909,19 +912,19 @@ Steps:
 6. pseudorandom_pad = HKDF-Expand(RwdU, concat(nonce, "Pad"), len(pt))
 7. auth_key = HKDF-Expand(RwdU, concat(nonce, "AuthKey"), Nk)
 8. exporter_key = HKDF-Expand(RwdU, concat(nonce, "ExporterKey", nonce), Nk)
-9. t' = HMAC(auth_key, concat(nonce, ct, concat(response.pkS, aad))), where aad is
-application-specific additional associated data.
-10. If !CT_EQUAL(t, t'), raise DecryptionError
-11. pt = xor(ct, pseudorandom_pad)
-12. C = DeserializeCredentials(pt)
-13. Output C, exporter_key
+9. auth_data = response.pkS
+10. t' = HMAC(auth_key, concat(nonce, ct, concat(response.pkS, aad)))
+11. If !CT_EQUAL(t, t'), raise DecryptionError
+12. pt = xor(ct, pseudorandom_pad)
+13. C = DeserializeCredentials(pt)
+14. Output C, exporter_key
 ~~~
 
 [[RFC editor: please change "RFCXXXX" to the correct number before publication.]]
 
 As in the registration phase, applications MUST authenticate pkS; secrecy of pkS is
 optional. If an application requires secrecy of pkS, this value SHOULD be omitted
-from the HMAC computation (step 9).
+from auth_data (step 9).
 
 ## Exporter Keys {#exporter-usage}
 
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -668,6 +668,8 @@ description of this encoding.
 struct {
     opaque data<0..2^16-1>;
     opaque pkS<0..2^16-1>;
+    CredentialType secret_types<2..254>;
+    CredentialType cleartext_types<2..254>;
 } RegistrationResponse;
 ~~~
 
@@ -721,6 +723,12 @@ Steps:
 ~~~
 CreateRegistrationResponse(request, pkS)
 
+Parameters:
+- secret_credentials_list, a list of CredentialType values clients should include
+ in the secret_credentials list of their Credentials structure
+- cleartext_credentials_list, a list of CredentialType values clients should include
+ in the cleartext_credentials list of their Credentials structure
+
 Input:
 - request, a RegistrationRequest structure
 - pkS, the server's public key
@@ -734,7 +742,8 @@ Steps:
 2. M = Deserialize(request.data)
 3. Z = Evaluate(kU, M)
 4. data = Z.encode()
-5. Create RegistrationResponse response with (data)
+5. Create RegistrationResponse response with
+     (data, pkS, secret_credentials_list, cleartext_credentials_list)
 6. Output (response, kU)
 ~~~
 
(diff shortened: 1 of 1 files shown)


hugokraw:
Hi Chris, I have several comments/corrections for PR #36.
Let me know if there is a better way to communicate such things in github
rather than as an email.

Here is a list (items marked by *)

Let me know if you have questions or need more clarity on these issues.


* Question (pure curiosity): What was the functionality of the "context"
value
what was context

* Line 642: Envelope is defined as an encryption of a Credentials
structure, but
  encryption is only required for skU. Maybe "an authenticated encoding"
instead
  of encryption, since everything in it requires authentication.

(same appears in line 821)

* params, the MHF parameters established out of band

Shouldn't we allow for params to be communicated by the server to client
(this
would be another potential element of Envelope)

* Calculation of keys using HKDF-Expand (lines 720..., also 891...)

Here is how the calculation should look:

pseudorandom_pad = HKDF-Expand(Key=RwdU, info = nonce | "Pad", len(pt))
auth_key         = HKDF-Expand(Key=RwdU, info = nonce | "AuthKey", Na)
exporter_key     = HKDF-Expand(Key=RwdU, info = nonce | "ExporterKey", Ne)

The keywords Key= and info= are not needed. RFC 5869 defines HKDF-Expand
with
three parameters corresponding to key, info and length, so just writing
pseudorandom_pad = HKDF-Expand(RwdU, nonce | "Pad", len(pt))
should be enough
(I write the name of the fields to clarify the semantics but this is not
needed
for specification).

Btw, I prefer "nonce" or "Nonce" better than n.

Also: Is Ne (the length of exporter_key) defined?
We can use the same length for auth_key and exporter_key - they should be
HashLen where this denotes the output length from the hash underlying HKDF
(or
HMAC)

* Encryption and aad

You define (line 718)
  pt = SerializeCredentials(C)
and then (line 723)
  ct = xor(pt, pseudorandom_pad)
which means you are mandating the encryption of all fields in Credentials.
I don't think we want to do that.
Some people (like Kevin and I got this feedback from AWS too) want to keep
pkS
unencrypted so we should accommodate that.
I also find the text in line 733-736 confusing.
Shouldn't the SerializeCredentials get as input an indication of what
Credentials fields are included and which are part of pt and which of aad?

In particular I don't like to leave this to the application:
  "Instantiations of OPAQUE MUST specify how aad is constructed and
serialized."
Shouldn't the OPAQUE client be standardized independently of the
applications?

> Is it ok to leave it up to the application? Don't you want to build
generic
clients? Can't the SerializeCredentials function have as input the
encrypted and
authenticated elements?

Btw, you use the notation pk(skU) but this is denoted by the simple pkU

* Important correction

Where it says (line 724):

t = HMAC(auth_key, concat(ct, aad)),  where aad is application-specific

it should be

t = HMAC(auth_key, concat(n, ct, aad)),  where aad is application-specific

the nonce needs to be authenticated (it's my fault, I overlooked it in the
text
I sent you).

* line 896 Abort(): should one define an error message for it?

chris-wood:
> * Question (pure curiosity): What was the functionality of the "context" value what was context 

It was meant to link requests/responses together. But that's something that can be handled by the instantiation (OPAQUE-EA in this case).

> * Line 642: Envelope is defined as an encryption of a Credentials structure, but encryption is only required for skU. Maybe "an authenticated encoding" instead of encryption, since everything in it requires authentication. (same appears in line 821) 

That works for me -- I'll use that text.

> * params, the MHF parameters established out of band Shouldn't we allow for params to be communicated by the server to client (this would be another potential element of Envelope) 

I don't think so. Clients can choose to harden the password using whatever parameters they wish, right? Why does the server need to have any input here?

> * Calculation of keys using HKDF-Expand (lines 720..., also 891...) Here is how the calculation should look: pseudorandom_pad = HKDF-Expand(Key=RwdU, info = nonce | "Pad", len(pt)) auth_key = HKDF-Expand(Key=RwdU, info = nonce | "AuthKey", Na) exporter_key = HKDF-Expand(Key=RwdU, info = nonce | "ExporterKey", Ne) The keywords Key= and info= are not needed. RFC 5869 defines HKDF-Expand with three parameters corresponding to key, info and length, so just writing pseudorandom_pad = HKDF-Expand(RwdU, nonce | "Pad", len(pt)) should be enough (I write the name of the fields to clarify the semantics but this is not needed for specification). 

To be clear, you're just suggesting removing the argument labels ("key="), right? I don't see any difference between concat(nonce, "Pad") and concat("Pad", nonce). 

> Btw, I prefer "nonce" or "Nonce" better than n. Also: Is Ne (the length of exporter_key) defined? We can use the same length for auth_key and exporter_key - they should be HashLen where this denotes the output length from the hash underlying HKDF (or HMAC) 

Yep, Ne is the length of the exporter key. It's a parameter to the functions where it's used. I folded Na and Ne together. I'll specify a value for this in a separate PR.

> * Encryption and aad You define (line 718) pt = SerializeCredentials(C) and then (line 723) ct = xor(pt, pseudorandom_pad) which means you are mandating the encryption of all fields in Credentials. I don't think we want to do that. Some people (like Kevin and I got this feedback from AWS too) want to keep pkS unencrypted so we should accommodate that. I also find the text in line 733-736 confusing. Shouldn't the SerializeCredentials get as input an indication of what Credentials fields are included and which are part of pt and which of aad? In particular I don't like to leave this to the application: "Instantiations of OPAQUE MUST specify how aad is constructed and serialized." Shouldn't the OPAQUE client be standardized independently of the applications?

I don't know of a good way to specify this in the pseudocode without a mess of branches. I think the default case should be to encrypt pkS, and we can specify variants (moving pkS out of Credentials and into aad, for example) in the text. But we can also flip that around and make the default case be to only authenticate pkS, with text suggesting it may optionally be encrypted. Would that work?

> Is it ok to leave it up to the application? Don't you want to build generic clients? Can't the SerializeCredentials function have as input the encrypted and authenticated elements? 

I'm not sure what you mean here. By leaving it up to applications, we are allowing generic clients to use this as needed.

> Btw, you use the notation pk(skU) but this is denoted by the simple pkU 

That works. :-)

> * Important correction Where it says (line 724): t = HMAC(auth_key, concat(ct, aad)), where aad is application-specific it should be t = HMAC(auth_key, concat(n, ct, aad)), where aad is application-specific the nonce needs to be authenticated (it's my fault, I overlooked it in the text I sent you). 

Whoops -- yes, will fix!

>* line 896 Abort(): should one define an error message for it?

I replaced this with "raise DecryptionError", to make it a bit more clear.

chris-wood:
@hugokraw -- I updated this based on your feedback above. Everything should be accounted for, but please let me know if not!

chris-wood:
@hugokraw, I pushed one more update that separates application AAD and the authentication-only data in credentials. Hopefully this is more clear!

hugokraw:
On Wed, Jun 17, 2020 at 10:43 PM Christopher Wood <notifications@github.com>
wrote:

>
>    - Question (pure curiosity): What was the functionality of the
>    "context" value what was context
>
> It was meant to link requests/responses together. But that's something
> that can be handled by the instantiation (OPAQUE-EA in this case).
>
>
>    - Line 642: Envelope is defined as an encryption of a Credentials
>    structure, but encryption is only required for skU. Maybe "an authenticated
>    encoding" instead of encryption, since everything in it requires
>    authentication. (same appears in line 821)
>
> That works for me -- I'll use that text.
>
>
>    - params, the MHF parameters established out of band Shouldn't we
>    allow for params to be communicated by the server to client (this would be
>    another potential element of Envelope)
>
> I don't think so. Clients can choose to harden the password using whatever
> parameters they wish, right? Why does the server need to have any input
> here?
>
The question is how the client knows at login time what parameters it has
to use. The client machine may be running OPAQUE instances with different
servers and users that have different parameters. In such a case, the
server needs to store these parameters (that were set a password
registration phase)  and send them to the client.


>    - Calculation of keys using HKDF-Expand (lines 720..., also 891...)
>    Here is how the calculation should look: pseudorandom_pad =
>    HKDF-Expand(Key=RwdU, info = nonce | "Pad", len(pt)) auth_key =
>    HKDF-Expand(Key=RwdU, info = nonce | "AuthKey", Na) exporter_key =
>    HKDF-Expand(Key=RwdU, info = nonce | "ExporterKey", Ne) The keywords Key=
>    and info= are not needed. RFC 5869 defines HKDF-Expand with three
>    parameters corresponding to key, info and length, so just writing
>    pseudorandom_pad = HKDF-Expand(RwdU, nonce | "Pad", len(pt)) should be
>    enough (I write the name of the fields to clarify the semantics but this is
>    not needed for specification).
>
> To be clear, you're just suggesting removing the argument labels ("key="),
> right? I don't see any difference between concat(nonce, "Pad") and
> concat("Pad", nonce).
>
Yes, that is what I was suggesting but note that the change from what you
had before is much more significant than that. You had
HKDF-Expand(salt=n, IKM=RwdU, "Pad", len(pt))
which uses the nonce as salt, but only HKDF-Extract uses salt, not Expand.

Btw, I prefer "nonce" or "Nonce" better than n. Also: Is Ne (the length of
> exporter_key) defined? We can use the same length for auth_key and
> exporter_key - they should be HashLen where this denotes the output length
> from the hash underlying HKDF (or HMAC)
>
> Yep, Ne is the length of the exporter key. It's a parameter to the
> functions where it's used. I folded Na and Ne together. I'll specify a
> value for this in a separate PR.
>
>
>    - Encryption and aad You define (line 718) pt =
>    SerializeCredentials(C) and then (line 723) ct = xor(pt, pseudorandom_pad)
>    which means you are mandating the encryption of all fields in Credentials.
>    I don't think we want to do that. Some people (like Kevin and I got this
>    feedback from AWS too) want to keep pkS unencrypted so we should
>    accommodate that. I also find the text in line 733-736 confusing. Shouldn't
>    the SerializeCredentials get as input an indication of what Credentials
>    fields are included and which are part of pt and which of aad? In
>    particular I don't like to leave this to the application: "Instantiations
>    of OPAQUE MUST specify how aad is constructed and serialized." Shouldn't
>    the OPAQUE client be standardized independently of the applications?
>
> I don't know of a good way to specify this in the pseudocode without a
> mess of branches. I think the default case should be to encrypt pkS, and we
> can specify variants (moving pkS out of Credentials and into aad, for
> example) in the text. But we can also flip that around and make the default
> case be to only authenticate pkS, with text suggesting it may optionally be
> encrypted. Would that work?
>
Why can't you have two components in Credentials: pt and aad, and the
application chooses which fields to map to each one? I understand that what
you propose is simpler but it also requires customizing the client to
different applications. In the case where the client is "written" for a
specific application, that's fine, but if you have a situation where the
same client interacts with different applications/servers then things
become problematic.

Is it ok to leave it up to the application? Don't you want to build generic
> clients? Can't the SerializeCredentials function have as input the
> encrypted and authenticated elements?
>
> I'm not sure what you mean here. By leaving it up to applications, we are
> allowing generic clients to use this as needed.
>
I am thinking of a web browser running OPAQUE (as a client) with different
servers that may have different policies regarding what is
included/encrypted/authenticated. So you want a standardized way for the
server to  communicate the structure of the envelope to the browser.

chris-wood:
> The question is how the client knows at login time what parameters it has to use. The client machine may be running OPAQUE instances with different servers and users that have different parameters. In such a case, the server needs to store these parameters (that were set a password registration phase)  and send them to the client.

Ah, okay, good point. I was assuming the application was "fixed," e.g., a mobile app in which the parameters are baked into the code. But that might not always be the case. I'll just revert this for now.

> Why can't you have two components in Credentials: pt and aad, and the application chooses which fields to map to each one? I understand that what you propose is simpler but it also requires customizing the client to different applications. In the case where the client is "written" for a specific application, that's fine, but if you have a situation where the same client interacts with different applications/servers then things become problematic.

Sent an email to discuss offline!

hugokraw:
I think the difference in our approaches/expectations is that you think of
OPAQUE for use inside a standalone application that has full control of the
client and I think more about a "universal" mechanism, e.g., a standardized
browser-based client.

draft-irtf-cfrg-opaque.md
```diff
@@ -615,44 +515,88 @@ struct {
...
+struct {
+  CredentialExtension secret_credentials<1..2^16-1>;
+  CredentialExtension cleartext_credentials<0..2^16-1>;
 } Credentials;
 ~~~
 
-skU
-: An encoded private key.
+secret_credentials
+: Application credentials which require secrecy and authentication.
```
hugokraw, line 568: I prefer OPAQUE credentials than application credentials

draft-irtf-cfrg-opaque.md
```diff
@@ -615,44 +515,88 @@ struct {
...
 
 Additionally, OPAQUE makes use of an additional structure `Credentials` to store
-user (client) credentials. It is structured as follows. Each public and
-private key value is an opaque byte string, specific to the AKE protocol in
-which OPAQUE is instantiated. For example, if used as raw public keys for
-TLS 1.3 {{?RFC8446}}, they may be RSA, DSA, or ECDSA keys as per {{?RFC7250}}.
+user (client) credentials. A `Credentials` structure consists of secret and
+cleartext `CredentialExtension` values. Each `CredentialExtension` indicates
+the type of extension and carries the raw bytes. This specification includes
+extensions essential for OPAQUE, including:
```
hugokraw, line 528: I would not call all of them essential, just skU and pkS are.

Do we want to restrict to just these five or leave further extensibility in case it is needed (e.g., we were discussing the hardening parameters as a possible addition to the envelope).

draft-irtf-cfrg-opaque.md
```diff
@@ -615,44 +515,88 @@ struct {
...
 
-skU
-: An encoded private key.
+secret_credentials
+: Application credentials which require secrecy and authentication.
 
-pkU
-: An encoded public key corresponding to private_key. This field is optional.
+cleartext_credentials
+: Application credentials which require authentication.
```
hugokraw, line 571: Applications credentials that require authentication but not secrecy

draft-irtf-cfrg-opaque.md
```diff
@@ -615,44 +515,88 @@ struct {
...
 
-idS
-: An identity of the server. This is typically a domain name, e.g., example.com.
-See {{SecIdentities}} for information about this identity. This field is optional.
+~~~
+struct {
+  opaque nonce[Nn];
+  opaque ct<1..2^16-1>;
+  opaque auth_data<0..2^16-1>;
+  opaque tag<1..2^16-1>;
```
hugokraw, line 585: maybe auth_tag instead of just tag? (not too important)

draft-irtf-cfrg-opaque.md
chris-wood: I noted that skU and pkS are mandatory. We might want to leave room for other things. The hardening parameters might be one of them, if they aren't transmitted at the application layer.

draft-irtf-cfrg-opaque.md
```diff
@@ -830,20 +763,31 @@ Steps:
...
 4. RwdU = Harden(y, params)
-5. credentials_key = HKDF(salt=0, IKM=RwdU, "CredentialsKey", Nk)
-6. exporter_key = HKDF(salt=0, IKM=RwdU, "ExporterKey", Nk)
-7. n = random(Nn) # generate a fresh nonce randomly
-8. Credentials C with (skU=skU, pkS=response.pkS)
-9. (ct, t) = Seal(credentials_key, n, "", C)
-10. EnvU = concat(n, ct, t)
-11. Create RegistrationUpload upload with envelope value EnvU
-11. Output (upload, exporter_key)
+5. Create Credentials C according to server policy
```
hugokraw, line 769: How is server policy (what is encrypted and what is not in the envelope) communicated to the client? DO we need a special message (echoing the credentials structure) for that?

draft-irtf-cfrg-opaque.md
```diff
@@ -802,21 +731,25 @@ Steps:
...
 
 Parameters:
 - params, the MHF parameters established out of band
+- Nn, length of the key derivation nonce
+- Nk, length of the authentication and exporter keys
 
 Input:
 - IdU, an opaque byte string containing the user's identity
 - PwdU, an opaque byte string containing the user's password
+- aad, application-specific additional associated data
```
hugokraw, line 754: what is the use of aad (additional to auth_data)?

chris-wood: Any AAD specific to the application, really. It's an opaque slot.

draft-irtf-cfrg-opaque.md
```diff
@@ -830,20 +763,31 @@ Steps:
...
-11. Output (upload, exporter_key)
+5. Create Credentials C according to server policy
+6. pt = SerializeExtensions(C.secret_credentials)
+7. nonce = random(Nn)
+8. pseudorandom_pad = HKDF-Expand(RwdU, contact(nonce, "Pad"), len(pt))
+9. auth_key = HKDF-Expand(RwdU, contact(nonce, "AuthKey"), Nk)
+10. exporter_key = HKDF-Expand(RwdU, concat(nonce, "ExporterKey"), Nk)
+11. ct = xor(pt, pseudorandom_pad)
+12. auth_data = SerializeExtensions(C.cleartext_credentials)
+13. t = HMAC(auth_key, concat(nonce, ct, concat(auth_data, aad)))
```
hugokraw, line 777: Do we need aad? If so, where is it transmitted? It is not part of the envelope.

draft-irtf-cfrg-opaque.md
chris-wood: I figured it'd be best to omit that from this, since it's an application-specific detail. But we could probably have the server be crystal clear about this inline.

draft-irtf-cfrg-opaque.md
chris-wood: Yes, I think so.

draft-irtf-cfrg-opaque.md
```diff
@@ -830,20 +763,31 @@ Steps:
...
 
-The application may optionally include pk(skU), IdS, and IdU in the Credentials
-structure C at step 8.
+Applications MUST encrypt and authenticate skU, and MUST authenticate pkS.
+Secrecy of pkS is optional. If an application requires secrecy of pkS, this value
+SHOULD be included in the `Credentials.secret_credentials` list (step 5).
+Applications may optionally include pkU, IdU, or IdS in the
+`Credentials.secret_credentials` structure (step 5) if secrecy of these values
+is desired. Otherwise, if an application does not require secrecy for these values
+but does require authentication, they may be appended to `Credentials.cleartext_credentials`.
```
hugokraw, line 791: We need to specify that *all* values in EnvU require authentication, only secrecy is optional (except for skU for which it is mandatory)

chris-wood: I think that's already clear from context, no?

draft-irtf-cfrg-opaque.md
```diff
@@ -956,45 +895,51 @@ Output:
...
 3. Z = Evaluate(kU, M)
 4. data = Z.encode()
-5. Create CredentialResponse response with (request.context, data, EnvU)
+5. Create CredentialResponse response with (data, EnvU, pkS)
 6. Output (response, pkU)
 ~~~
 
-#### RecoverCredentials(IdU, PwdU, metadata, request, response)
+Applications which include pkS in the Credentials structure MAY omit
+pkS from the CredentialResponse.
```
hugokraw, line 931: pkS is a mandatory part of EnvU

draft-irtf-cfrg-opaque.md
```diff
@@ -1007,16 +952,28 @@ Steps:
 1. Z = Deserialize(response.data)
 2. N = Unblind(input.data_blind, Z)
 3. y = Finalize(PwdU, N, "RFCXXXX")
-4. RwdU = Harden(y, params)
-5. credentials_key = HKDF(salt=0, IKM=RwdU, "CredentialsKey", Nk)
-6. exporter_key = HKDF(salt=0, IKM=RwdU, "ExporterKey", Nk)
-7. (n, ct, t) = response.envelope
-8. C = Open(credentials_key, n, "", ct, t)
-9. Output C, exporter_key
+4. nonce = response.envelope.tag
```
hugokraw: why tag?

draft-irtf-cfrg-opaque.md
```diff
@@ -830,20 +763,31 @@ Steps:
...
-7. n = random(Nn) # generate a fresh nonce randomly
-8. Credentials C with (skU=skU, pkS=response.pkS)
-9. (ct, t) = Seal(credentials_key, n, "", C)
-10. EnvU = concat(n, ct, t)
-11. Create RegistrationUpload upload with envelope value EnvU
-11. Output (upload, exporter_key)
+5. Create Credentials C according to server policy
+6. pt = SerializeExtensions(C.secret_credentials)
+7. nonce = random(Nn)
+8. pseudorandom_pad = HKDF-Expand(RwdU, contact(nonce, "Pad"), len(pt))
```
hugokraw, line 772: Make reference to RFC 5869 about the roles of the three inputs to HKDF-Expand, namely, key, info, and output length.

draft-irtf-cfrg-opaque.md
chris-wood: (This list is meant to be extended, basically.)

draft-irtf-cfrg-opaque.md
chris-wood: I removed this left-over text.

draft-irtf-cfrg-opaque.md
chris-wood: Bug -- fixed!

draft-irtf-cfrg-opaque.md
```diff
@@ -1514,8 +1455,7 @@ Note that if the same pair of user identity IdU and value alpha is received
 twice by the server, the response needs to be the same in both cases (since
 this would be the case for real users).
 For protection against this attack, one would apply the encryption function in
-the construction of EnvU ({{SecEnvU}}) to all the key material in EnvU,
-namely, AOenv will be empty.
+the construction of EnvU to all the key material in EnvU, namely, aad will be empty.
```
hugokraw, line 1461: I guess that aad should be replaced here with cleartext_credentials?

draft-irtf-cfrg-opaque.md
chris-wood: I added a reference at the end of this algorithm.

draft-irtf-cfrg-opaque.md
chris-wood: Yep, good catch!

PR #38, opened 2020-06-18
Remove TLS 1.3 integration details.

### LINKED ISSUE #2: Drop TLS integration details
We should reference and rely on [1] when it's updated.

[1] https://tools.ietf.org/html/draft-sullivan-tls-opaque-00

chris-wood:
Closes #2.

This removes TLS 1.3 integration details, but keeps the SIGMA-I content. (TLS 1.3 details will go in the TLS document.)


chris-wood:
I love that line :-D

PR #40, opened 2020-07-09
Exporter -> export, adding HKDF-Extract around Harden call

kevinlewi:
We should wrap the Harden() call with an HKDF-Extract call, since there is no guarantee that Harden will output pseudorandom bits

Also changed "exporter key" to read simply as "export key"


kevinlewi:
@chris-wood

PR #47, opened 2020-08-24
Adopt the TLS 1.3 key schedule for better separation.

chris-wood:


draft-irtf-cfrg-opaque.md
```diff
@@ -1052,7 +1052,34 @@ for binding between the underlying OPRF protocol messages and the KE session.
...
+HKDF-Expand-Label(Secret, Label, Context, Length) =
+  HKDF-Expand(Secret, HkdfLabel, Length)
+~~~
+
+Where HkdfLabel is specified as:
+
+~~~
+struct {
+   uint16 length = Length;
+   opaque label<7..255> = "tls13 " + Label;
```
hugokraw, line 1070: "OPAQUE" instead of "tls13" I guess (use upper case OPAQUE to distinguish from the opaque type.

draft-irtf-cfrg-opaque.md
```diff
@@ -1052,7 +1052,34 @@ for binding between the underlying OPRF protocol messages and the KE session.
...
+   opaque context<0..255> = Context;
+} HkdfLabel;
+
+Derive-Secret(Secret, Label, Transcript) =
+    HKDF-Expand-Label(Secret, Label, Hash(Transcript), Hash.length)
```
hugokraw, line 1075: What I call Ke2, Ke3  is the equivalent of the handshake write keys in TLS 1.3, and Km2, Km3 are the equivalent of Finished keys. What I call SK is a key from which traffic write keys are derived. 
So the equivalent to TLS would be to use  Derive_Secret to output handshake_traffic_secret and application_traffic_secret and then define the write_key derivations as in Section 7.3 of TLS 1.3 and the Finished key derivation as in Section 4.4 and 4.4.4.
If you want to simplify and keep it the way you defined it now, note that Ke2 and Ke3 cannot use hash.length but would have a length that depends on the ciphersuite's AEAD.

draft-irtf-cfrg-opaque.md
chris-wood: Copy-paste bug :) Will fix!

draft-irtf-cfrg-opaque.md
chris-wood: Ah! Thanks for clarifying. I split the schedule into two parts: one for handshake secrets (Ke2, Ke3, Km2, Km3) and output AKE secrets (SK). All handshake keys, including the encryption and Finished/MAC keys, are derived from the client/server handshake secrets. What do you think?

hugokraw: Here is a simplification

  HKDF-Extract(salt=0, IKM) = main_secret
      |
      +--> Derive-Secret(., "c hs secret", info) = client_handshake_secret
      |
      +--> Derive-Secret(., "s hs secret", info) = server_handshake_secret
      |
      +--> Derive-Secret(., "c sk secret", info) = client_session_key
      |
      +--> Derive-Secret(., "s sk secret", info) = server_session_key
      |
      v

And it can even be simplified further by deriving a single handshake_secret and single session_key and do the separation client/server in further derivation.

It really depends how close to TLS 1.3 you want to be.  For example, do we want to stop at the session_key derivation or also derive the "write keys" for the AEAD to protect subsequent traffic as TLS 1.3 does?

hugokraw: Btw, getting too close to TLS 1.3 in names (e.g. client_handshake_secret), we may need to add "OPAQUE" as a label. Or maybe this is only done when actually integrating with TLS.

chris-wood: This is quite nice -- I'll go with this. I think it's fine to leave further key derivation to the protocol in question, at least for now.

chris-wood: Yeah, I think it's fine to reuse the names here. I applied the simplification you suggested (compressed to a single handshake and session secret), and then derived all handshake secrets (MAC and enc keys) from the handshake_secret.

PR #48, opened 2020-08-27
Fix the nonce length (Nn) to 32 bytes.

chris-wood:
Making this a parameter seems like too much flexibility, so I suggest we be opinionated and remove it.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -561,7 +561,7 @@ It is encoded as follows.
 
 ~~~
 struct {
-  opaque nonce[Nn];
+  opaque nonce[32];
   opaque ct<1..2^16-1>;
   opaque auth_data<0..2^16-1>;
   opaque auth_tag<1..2^16-1>;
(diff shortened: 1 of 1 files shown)


chris-wood:
> that in the future someone will want to steal a few bytes here and there from the nonce and then... :- )

Hah! :-)

> Anyway, I would keep the Nn but I am ok if you feel strongly about getting rid of it.

My take is that an OPAQUE configuration can simply specify an OPRF, hash function (for HMAC and HKDF), and MHF, without any additional parameters. 

How about I merge this with a note saying we could make this a parameter if flexibility is desired, and then as the larger group for feedback?

hugokraw:
I'm fine with the 32 bytes. There is enough room to steal from it if needed. I can see it as more probable that someone will want to save in EnvU size than that 32 bytes will not be enough.
Btw, you also need an AKE to specify OPAQUE in addition to the components you mention.

chris-wood:
> Btw, you also need an AKE to specify OPAQUE in addition to the components you mention.

Yep!

PR #49, opened 2020-08-28
Add a hash function dependency

chris-wood:
And use it to control the auth_key and exporter_key size(s).


chris-wood:
Yeah, I agree that it's overkill. I'll mark this as an OPEN ISSUE to consider fixing in the OPRF draft. (There was no reason to use SHA-512, other than being conservative, I think.)

PR #61, opened 2020-09-27
Update draft-irtf-cfrg-opaque.md

hugokraw:
I made some minir edits, nothing worth highlighting I think.
Two issues that we may want to address before submitting are:

1. We say IBM has a patent on HMQV. Is that enough (for now at least) or do we need to have some formal IP notice?

2. Section 6.3 (called from section 3.2 and 4.2.1) refers to SP800-56A Sec 5.6.2.3.4 which seems to be phrased for ephemeral keys while in the context of Sec 3.2 we talk about static ones. More importantly, the NIST document omits any tests of membership as they assume/require cofactor clearing. Is this something we want to assume/require ourselves? I thought we would "outsource" these issues to documents describing the curves themselves as the choice between explicit membership tests, prime order tests, cofactor operations, etc. may depend on the specific curve. If you do not want to get into these issues now (I am happy not to), we should still mention membership tests in Section 6.3. 

Once we are done with these two issues, we can submit.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1348,7 +1348,9 @@ is as specified in Section 5.6.2.3.4 of {{keyagreement}}. This includes
 checking that the coordinates are in the correct range, that the point
 is on the curve, and that the point is not the point at infinity.
 Additionally, validation MUST ensure the Diffie-Hellman shared secret is
-not the point at infinity.
+not the point at infinity. For X25519 and X448, validation is as described in
+{{?RFC7748}}. In particular, where applicable, endpoints MUST check whether
+the Diffie-Hellman shared secret is the all-zero value and abort if so.
 
 ## User authentication versus Authenticated Key Exchange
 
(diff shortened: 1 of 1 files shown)


chris-wood:
> We say IBM has a patent on HMQV. Is that enough (for now at least) or do we need to have some formal IP notice?

I think you can submit IPR here: https://datatracker.ietf.org/ipr/

> If you do not want to get into these issues now (I am happy not to), we should still mention membership tests in Section 6.3.

I think this is covered by "MUST validate the other party's public key(s) used for the execution of OPAQUE", no? If not, what would you change?

hugokraw:
Regarding:

> >If you do not want to get into these issues now (I am happy not to), we should still mention membership tests in Section 6.3.

> I think this is covered by "MUST validate the other party's public key(s) used for the execution of OPAQUE", no? If not, what would you change?

The problem is that we refer to the NIST document that does not include membership tests as they assume cofactor clearing which we don't assume. 
Note that we say explicitly:

epkU, epkS are Diffie-Hellman ephemeral public keys chosen by user and server, respectively, which MUST be validated to be in the correct group (see {{validation}});

but then the membership test is lost in the reference to NIST.

chris-wood:
The NIST document pertains only to the NIST curves, which don't require cofactor clearing (h=1). The other curves may involve cofactor clearing, but I think that's covered by the public key validation text.

hugokraw:
Oh. Ok. I missed the (clear) point that this is an example that only pertains to the NIST curves. Sorry for the confusion. You can go ahead and submit. Thanks.

chris-wood:
Perfect -- thanks!

PR #70, opened 2020-10-21
Add a preliminary Sage implementation with many TODOs

chris-wood:
This gets us started towards test vectors, with something that's "as close" to the specification as possible. There's definitely lots of ways this can be improved, including adding proper configuration and agility support, supporting different credentials in client and server, implementing a proper password file, etc. If this generally seems useful, we can add these missing pieces and then wrap it with code that generates test vectors from a variety of flows. 

Note that this *does not* have AKE integration. It only implements the OPAQUE core protocol. For the specification, I wonder if it makes sense to include test vectors for both the core protocol and AKE instantiations. 

@kevinlewi, please have a look!


chris-wood:
Closing for now -- I'll re-open when the 3DH implementation is in place.

Issue #72, opened 2020-10-22
Why does FinalizeRequest take in idU as a parameter?


kevinlewi:
According to [this](https://github.com/cfrg/draft-irtf-cfrg-opaque/blame/master/draft-irtf-cfrg-opaque.md#L651), idU is used to construct FinalizeRequest, despite it only being applicable if the application decides to include it in the Credentials struct: > "Applications may optionally include
> `pkU`, `idU`, or `idS` in the `Credentials.cleartext_credentials` structure, or in
> `Credentials.secret_credentials` if secrecy of these values is desired."

But if this is the case, then why not also make idS and pkU parameters to FinalizeRequest as well?

Same goes for RecoverCredentials. Either they should take in all of these optional parameters as inputs, or they should not include any of them.

chris-wood:
Proposal: remove application-specific credential information from the parameter list, and call out to the application to build `Credentials` based on the server's `secret` and `cleartext` list.

Issue #65, opened 2020-10-19
Omitting idU parameter in RegistrationRequest and CredentialRequest messages


kevinlewi:
The idU parameter is not used by the server when receiving the `RegistrationRequest` and `CredentialRequest` messages from the client. Can this parameter be omitted from the specification of these messages?

In theory, the server does need some "username" equivalent in order to look up the client's record, so this information does need to be sent, but I don't think it should be included in the OPAQUE messages, since it has no function within OPAQUE other than to set/retrieve the appropriate password file.

Presumably, any wrapper protocol which uses OPAQUE will not want to look into the bytes of the OPAQUE messages in order to figure out what the username is.

chris-wood:
Well, it needs to be included in the transcript for the AKEs, I think, so that's why it's part of the `CredentialRequest` message.

kevinlewi:
The presence of idU is dependent on the AKE protocol used, and so even though it may get sent over as the first message of the AKE, shouldn't this be controlled by the specification of the AKE, and not the OPRF component?

Also, this is only for `CredentialRequest`, since `RegistrationRequest` does not use idU. I would be happy with delegating the specifying of idU to the AKE component.

chris-wood:
> The presence of idU is dependent on the AKE protocol used, and so even though it may get sent over as the first message of the AKE, shouldn't this be controlled by the specification of the AKE, and not the OPRF component?

That's true. One thing I have in my implementation notes is whether we ought to include *both* `idU` and `pwdU` in the OPRF input, rather than just `pwdU`. Given that `kU` is unique per `idU`, this probably doesn't make a meaningful change in security, though I wonder if it's just cleaner aesthetically?

I could go either way, I guess. I'm curious to hear what @hugokraw thinks.

kevinlewi:
>  One thing I have in my implementation notes is whether we ought to include both idU and pwdU in the OPRF input, rather than just pwdU.

That is certainly a valid consideration, but regardless of whether or not this is changed, I believe we can still omit idU from the `RegistrationRequest` and `CredentialRequest` messages, since the server doesn't use them (except for in the AKE component potentially).

chris-wood:
True, though I wonder if that just complicates things. (I don't disagree with the proposal. I just need to think about it more.)

hugokraw:
[...]

PR #73, opened 2020-10-22
Omitting idU from RegistrationRequest and CredentialRequest

kevinlewi:
This removes the inclusion of idU from RegistrationRequest and CredentialRequest (see #65)

I also noticed that the functions `RecoverCredentials` and `FinalizeRequest` should either not take idU, or should take idU, idS, and pkU as parameters. In this PR I simply made RecoverCredentials also take idU as a parameter, but I opened #72 to discuss this further.


kevinlewi:
@chris-wood : This should be ready to review

PR #76, opened 2020-10-28
Remove the performance considerations section.

### LINKED ISSUE #5: Reconsider "cost" details section
The "cost" details are written assuming a traditional DH-based KE protocol. Should we try to generalize this, e.g., to cover KEM-based protocols? Moreover, the cost details might not age well as TLS 1.3 develops and possibly moves away from signatures, via, e.g., semi-static DH. Or we might just remove this section entirely.

chris-wood:
Closes #5.

It was somewhat stale, and I'm not sure it adds much to the document.

cc @kevinlewi


kevinlewi:
This is fine by me

Issue #74, opened 2020-10-27
Clarifying optional versus required parameters in the AKE section


kevinlewi:
In this [section on AKE](https://github.com/cfrg/draft-irtf-cfrg-opaque/blame/master/draft-irtf-cfrg-opaque.md#L1008-L1014), we should clear up exactly which parameters are required and which are optional. Then we can also work towards defining a wire format for these AKEs if necessary.

A couple of concrete questions:
1) Can we simply set idU = static pkU and idS = static pkS?
2) Can we remove mention of all of the following: info1*, info2*, Einfo2*, info3*, Einfo3*?

chris-wood:
> 1. Can we simply set idU = static pkU and idS = static pkS?

This had been discussed in the past, and I think it depends on whether or not there exists a secure binding between the public key(s) and corresponding identity. (And that seems mostly external to the protocol.) I need to think about this more. @hugokraw probably also has opinions. :-)

> 2. Can we remove mention of all of the following: info1*, info2*, Einfo2*, info3*, Einfo3*?

These fields exist to make the AKE instantiation more generic, e.g., maybe `info1` contains the client's set of application protocols run on top of the AKE. I think we should remove them, since applications that need this flexibility should be using TLS. I view the 3DH and SIGMA-I variants as minimalist OPAQUE instantiations.

hugokraw:
Regarding point 1: In general, identities refer to logical entities on the basis of which policy decisions by the receiving party are made (do I want to talk to this identity, under what security level, which access rights, etc.) A public key may be used as an identity but most of the time it is not (e.g. , it is the subject in a certificate rather than the public key). In the case of OPAQUE,  using public keys may be more acceptable since those are fixed after registration and the client authenticates the server's public key via the password. I am not sure if this means that the client will never need to make decisions based on the server's logical identity but I do not see an obvious setting where this is the case. However, note that in the three protocols in the draft, the client does not use the user's public key. So if we are going to use it then we either send it as part of the authenticated information in the envelope or require the client to re-compute it from the user's private key at the cost of an exponentiation.

hugokraw:
Regarding point 2: We want these AKEs to be minimalistic but not useless. I do not see the case of any key exchange protocol that does not require sending some additional setting-specific information in its flows. Instead of having these fields named specifically, you may define some extensibility mechanism to transport additional information, but some mechanism for that must exist.

kevinlewi:
Thank you for the comments!

[...]

PR #81, opened 2020-11-10
Defining idU and idS in AKE section

kevinlewi:
Points mentioned after discussion from #74 

- Removes the sending of idU in KE1
- Adds clarification text around what idU and idS refers to in the AKE key derivation. By default, they will be taken from the idU and idS specified in the envelope. If they are not specified, then they are taken to be equal to pkU and pkS.
- Also clarify that if pkU is not in the envelope, then it must be computed from skU.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -20,7 +20,7 @@ author:
  -
     ins: K. Lewi
     name: Kevin Lewi
-    organization: Novi
+    organization: Novi Research
     email: lewi.kevin.k@gmail.com
  -
     ins: C. A. Wood
@@ -995,7 +995,7 @@ are performed in this group and represented here using multiplicative notation.
 
 OPAQUE with HMQV and OPAQUE with 3DH comprises:
 
-- KE1 = credential_request, nonceU, info1*, idU*, epkU
+- KE1 = credential_request, nonceU, info1*, epkU
 - KE2 = credential_response, nonceS, info2*, epkS, Einfo2*, MAC(Km2; transcript2),
 - KE3 = info3*, Einfo3*, MAC(Km3; transcript3)}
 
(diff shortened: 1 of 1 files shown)


chris-wood (approved):
This matches my understanding. @hugokraw, can you please have a look?

hugokraw:
I am ok with these changes, they reflect the conclusions of our discussion. 
I have a feeling that people will wonder about why are pkU, pkS set as default identities and in what cases they should or should not set different values for idU and idS.  Answering these question would require iterating some of the issues discussed in our thread (why identities are needed at all, what's their roles, etc.). Do you think there is some text you could write to explain some of these things and preempt some of the confusion? I would not dare to write such text myself :-)

kevinlewi:
> I am ok with these changes, they reflect the conclusions of our discussion.
> I have a feeling that people will wonder about why are pkU, pkS set as default identities and in what cases they should or should not set different values for idU and idS. Answering these question would require iterating some of the issues discussed in our thread (why identities are needed at all, what's their roles, etc.). Do you think there is some text you could write to explain some of these things and preempt some of the confusion? I would not dare to write such text myself :-)

@hugokraw Please take a look at the text that I have added to the "Security Considerations" section, also copied below:

```
## User and server identities

The user identity (idU) and server identity (idS) are optional parameters
which are left to the application to designate as monikers for the client
and server. If the application layer does not supply values for these
parameters, then they will be omitted from the creation of the envelope
during the registration stage. Furthermore, they will be substituted with
idU = pkU and idS = pkS during the authenticated key exchange stage.

The advantage to supplying a custom idU and idS (instead of simply relying
on a fallback to pkU and pkS) is that the client can then ensure that any
mappings between idU and pkU (and idS and pkS) are protected by the
authentication from the envelope. Then, the client can attempt to verify
that the idU and idS contained in its envelope matches the idU and idS
supplied by the server.

However, if this extra layer of verification is unnecessary for the
application, then simply leaving idU and idS unspecified is acceptable.
```
---

Let me know if you think we should edit this.

hugokraw:
I like it. Two edits:
- remove "attempt to"
- edit last sentence to: then simply leaving idU and idS unspecified, and using pkU and pkS instead, is acceptable.

kevinlewi:
Done. Thanks!

PR #86, opened 2020-11-16
Remove protocol messages.

### LINKED ISSUE #85: protocol messages
in the specification there is a part which defines `{#protocol-messages}`:
```
enum {
    registration_request(1),
    registration_response(2),
    registration_upload(3),
    credential_request(4),
    credential_response(5),
    (255)
} ProtocolMessageType;

struct {
    ProtocolMessageType msg_type;    /* protocol message type */
    uint24 length;                   /* remaining bytes in message */
    select (ProtocolMessage.msg_type) {
        case registration_request: RegistrationRequest;
        case registration_response: RegistrationResponse;
        case registration_upload: RegistrationUpload;
        case credential_request: CredentialRequest;
        case credential_response: CredentialResponse;
    };
} ProtocolMessage;
```
in our implementation at https://github.com/stef/libopaque we omitted implementing this, for the following reasons:
  * this is depending on how an application is using this protocol,  it might not be necessary to wrap protocol messages in such frames, for example because the app already has it's own framing, or different ways to communicate this.
  * parsing any kind of input from possible adversaries should be done higher up the stack, where possibilities are richer to do so in memory safe ways or by implementing language security mechanisms to do so.

libopaque does however implement the credential and envelope structures.

chris-wood:
Closes #85. 

cc @kevinlewi, @stef

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -403,7 +403,8 @@ Each public and private key value is an opaque byte string, specific to the AKE
 protocol in which OPAQUE is instantiated. For example, if used as raw public keys
 for TLS 1.3 {{?RFC8446}}, they may be RSA or ECDSA keys as per {{?RFC7250}}.
 
-The full `Credentials` encoding is as follows.
+The full `Credentials` encoding is as follows, described using TLS notation
+(see {{RFC8446}}, Section 3).
 
 ~~~
 enum {
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -368,55 +368,28 @@ We write `(kU, _) = KeyGen()` to denote use of this function for generating secr
 # Core Protocol {#protocol}
 
 OPAQUE consists of two stages: registration and authenticated key exchange.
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -368,55 +368,28 @@ We write `(kU, _) = KeyGen()` to denote use of this function for generating secr
...
 including: HMQV, 3DH, and SIGMA-I. {{I-D.sullivan-tls-opaque}} discusses integration with
 TLS 1.3 {{RFC8446}}.
 
-## Protocol messages {#protocol-messages}
+## Data types {#data-types}
 
-The OPAQUE protocol runs the OPRF protocol in two stages: registration and
-authenticated key exchange. A client and server exchange protocol messages in
-executing these stages. This section specifies the structure of these protocol
-messages using TLS notation (see {{RFC8446}}, Section 3).
```
kevinlewi, line 390: Perhaps we want to keep this sentence:

> This section specifies the structure of these protocol
messages using TLS notation (see {{RFC8446}}, Section 3).

Since it still applies for the remainder of the document

kevinlewi (approved):

PR #93, opened 2020-12-01
Move idU to CreateCredentialResponse parameter.

### LINKED ISSUE #88: id missing in CredentialRequest
in the section https://github.com/cfrg/draft-irtf-cfrg-opaque/blob/master/draft-irtf-cfrg-opaque.md#authenticated-key-exchange-messages CredentialRequest is defined as:

```
struct {
    opaque data<1..2^16-1>;
} CredentialRequest;
```
and right below that is this line:
> data : An encoded element in the OPRF group. See {{I-D.irtf-cfrg-voprf}} for a description of this encoding.
and in the https://github.com/cfrg/draft-irtf-cfrg-opaque/blob/master/draft-irtf-cfrg-opaque.md#createcredentialresponserequest section it is even referred to as:
```
1. (kU, envU, pkU) = LookupUserRecord(request.id)
```

even though trivial, it might make sense to include id in the CredentialRequest structure definition.

chris-wood:
The OPAQUE wrapper protocol is responsible for transmitting idU from
client to server. OPAQUE does not use it for anything other than password
file lookup.

Closes #88.

cc @kevinlewi, @stef

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -782,7 +782,9 @@ Steps:
 CreateCredentialResponse(request)
 
 Parameters:
-- idU, the identity associated with the user
+- kU, OPRF key associated with idU
+- envU, Envelope associated with idU
+- pkU, Public key associated with idU
 
 Input:
 - request, a CredentialRequest structure
@@ -792,10 +794,9 @@ Output:
 - pkU, public key of the user
 
 Steps:
-1. (kU, envU, pkU) = LookupUserRecord(idU)
-2. Z = Evaluate(kU, request.data)
-3. Create CredentialResponse response with (Z, envU)
-4. Output (response, pkU)
+1. Z = Evaluate(kU, request.data)
+2. Create CredentialResponse response with (Z, envU)
+3. Output (response, pkU)
 ~~~
 
 #### RecoverCredentials(pwdU, metadata, request, response)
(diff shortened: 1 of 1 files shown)


kevinlewi:
This looks good, but I do have a general comment that maybe we can move to discussion in an issue:

Would it make more sense to perhaps make kU, envU, pkU parameters to CreateCredentialResponse? The reason is because "LookupUserRecord" is not actually defined anywhere, and is only run in the wrapper protocol, so it may cause confusion.

chris-wood:
> Would it make more sense to perhaps make kU, envU, pkU parameters to CreateCredentialResponse? The reason is because "LookupUserRecord" is not actually defined anywhere, and is only run in the wrapper protocol, so it may cause confusion.

Yeah, I like that suggestion. Stand by while I make the change!

chris-wood:
@kevinlewi fixed!

kevinlewi:
Cool, looks! Minor nit: Shouldn't kU, envU, pkU be listed in the Input, not the Parameters (and also following an update to read as: "CreateCredentialResponse(request)" -> "CreateCredentialResponse(request, kU, envU, pkU))

chris-wood:
I opted to keep them as parameters since they're handled by the wrapper protocol and provided by context. Would it be more clear if they were inputs?

kevinlewi:
We can leave it to a separate change since it is probably a separate discussion, but I also noted that for "CreateRegistrationResponse", pkS is an input instead of a parameter. Not sure if that should change, but it seems like they should both be parameters or both be inputs.

Anyway, not a big deal either way, at least for now!

chris-wood:
Good catch -- moved to inputs!

kevinlewi:
Excellent

PR #95, opened 2020-12-03
Remove wrapper RequestMetadata type.

### LINKED ISSUE #94: Remove RequestMetadata type
This struct stores the scalar across Request/Finalize calls, which both happen on the client. There's no need to serialize or send this value on the wire anywhere, so we should just defer its representation to the underlying implementation. Given that the OPRF API is already a dependency we expose, I suggest we replace RequestMetadata with the actual type `Scalar`. 

While we're in here, we might also consider removing superfluous wrappers around OPRF types (like `RegistrationRequest` and `CredentialRequest`). 

@kevinlewi, what do you think?

chris-wood:
Closes #94.

cc @kevinlewi

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -307,7 +307,7 @@ operations, roles, and behaviors of OPAQUE:
 - Client (U): Entity which has knowledge of a password and wishes to authenticate.
 - Server (S): Entity which authenticates clients using passwords.
 - (skX, pkX): An AKE key pair used in role X; skX is the private key and pkX is
-the public key. For example, (skU, pkU) refers to U's private and public key.
+  the public key. For example, (skU, pkU) refers to U's private and public key.
 - kX: An OPRF private key used in role X. For example, kU refers to U's private OPRF
   key.
 - I2OSP and OS2IP: Convert a byte string to and from a non-negative integer as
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -796,18 +780,18 @@ Steps:
...
+RecoverCredentials(pwdU, blind, request, response)
 
 Parameters:
 - params, the MHF parameters established out of band
 - Nh, the output size of the Hash function
 
 Input:
 - pwdU, an opaque byte string containing the user's password
-- metadata, a RequestMetadata structure
+- blind, a RequestMetadata structure
```
kevinlewi, line 794: "a RequestMetadata structure" should be replaced with "an OPRF Scalar value"

draft-irtf-cfrg-opaque.md
```diff
@@ -517,33 +517,24 @@ See {{validation}} for more details.
...
-} RequestMetadata;
-~~~
-
-data_blind
-: An encoded OPRF scalar element.
-
-~~~
-struct {
-    opaque data<0..2^16-1>;
+    SerializedElement data;
```
kevinlewi, line 529: Where is SerializedElement defined? Perhaps a reference to an RFC #?

draft-irtf-cfrg-opaque.md
chris-wood: Missed this one -- thanks!

draft-irtf-cfrg-opaque.md
chris-wood: I clarified that this is an alias for a serialized group element, i.e., output of `Serialize()`.

kevinlewi (approved):
Looks good!

PR #101, opened 2020-12-17
Reference "Partitioning Oracle Attacks" paper.

### LINKED ISSUE #100: export_keys also need AE with RKR
since my implementation has been prominently featured in the "partitioning oracles paper" i was thinking that the export_key must also only be used with key-committing AE schemes. i think this should be emphasized where export_keys are mentioned in the draft.

chris-wood:
Closes #100.

cc @hugokraw, @stef

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -163,7 +163,7 @@ online)"
     seriesinfo: IEEE European Symposium on Security and Privacy
     date: 2016
 
-  PartioningOracleAttacks:
+  LGR20:
     title: Partitioning Oracle Attacks
     target: https://eprint.iacr.org/2020/1491.pdf
     author:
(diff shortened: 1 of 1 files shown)


hugokraw (commented):
Wait with this reference. Not sure if we need to touch on this point at all as there is no key robustness requirement for this specification. At most, it could be discussed in the security considerations just because the OPAQUE paper requires it. In this case we would say that while the paper requires it, this specification does not (in particular, partition attacks as those described in the Julia et al paper are irrelevant to this specification).

chris-wood:
> In this case we would say that while the paper requires it, this specification does not (in particular, partition attacks as those described in the Julia et al paper are irrelevant to this specification).

So, to be clear, would you just drop the reference from this PR? If so, we lose a pointer to the RKR definition, which I was hoping to outsource to somewhere else.

hugokraw:
I think the following text in the security considerations addresses this issue.

We note that the analysis of OPAQUE from {{OPAQUE}} requires the authenticated encryption scheme used to produce envU to have a special property called random key-robustness (or key-committing). This specification enforces this property by utilizing encryption-then-hmac in the construction of envU. There is no option to use another authenticated-encryption scheme with this specification.  (Deviating from the key-robustness requirement may open the protocol to attacks, e.g., {{Len}}.) We remark that export_key for authentication and/or encryption requires no special properties from the authentication/encryption schemes as long as exporter_key is used only after the EnvU is validated (i.e., after the HMAC in RecoverCredentials passes verification).

chris-wood:
That's much better. I'll update this PR with that suggestion and then close out this issue.

Issue #99, opened 2020-12-04
Simplifying the options for the credentials list in the envelope


kevinlewi:
This issue is related to #87 and #90, in which the complexities of the setting of the envelope have come up.

One problem I foresee happening with allowing the server to choose what parameters should be optional, [which go in secret_credentials, which go in cleartext_credentials, etc.](https://github.com/cfrg/draft-irtf-cfrg-opaque/blame/master/draft-irtf-cfrg-opaque.md#L438-L442) is that it becomes cumbersome to manage exhaustive test vectors that would test all of the edge cases (combinations of these settings). And if we choose to only provide test vectors for some default setting, we run the risk of an implementation in the wild incorrectly implementing support for a non-default setting that we did not provide test vectors for.

To remedy this, one approach we could take would be to allow for only a small (maybe 2?) set of options for organizing the credentials in the envelope. Perhaps a "base mode" and a "custom identifier mode", and we could spend a single byte to dictate the mode for the envelope (similar to the VOPRF RFC with the byte representing the base mode or verifiable mode). Base mode would be some default, and "custom identifier mode" would support a custom idU and idS being included as secret_credentials.

If this sounds good, then now we just need to decide what modes we will support, and what each mode will look like.

chris-wood:
I like this idea. Given that the number of credential types is likely quite small, I'm in favor of simplifying things and offering two modes here. For the "custom identifier mode," either idU or idS could be nil, if desired, right? Or would we require that they both non-nil?

kevinlewi:
I think we should allow for nil idU and idS, because it is not easy to check/enforce that idU and idS, and even if it were easily checkable, it is still an additional check that the OPAQUE implementation would have to do, which increases complexity.

bytemare:
Both idu and ids are used in the "info" / "transcript" (naming is not consistent) parameter in HKDF. Having them _nillable_ means the HKDF only relies on the nonces. But I don't know if that's a problem.

kevinlewi:
The default behavior (base mode) is for them to be set to pkU and pkS. Only in customIdentifier mode can they be set to arbitrary strings. We discussed this a bit -- and we decided to allow null values to be set, since it is up to the wrapper protocol to dictate what should represent valid identities for idU and idS. Even if we were to disallow null values, they could still be set to 1-byte values... so the best we can do is to advise for defaulting to base mode unless the implementer really wants to use the custom identifier mode in this manner.

[...]

PR #102, opened 2020-12-18
Changing envelope definition to  support two fixed modes

kevinlewi:
See #99.

This defines two modes, `base` and `customIdentifier` for the envelope. The credentials are incorporated in the `SecretCredentials` and `CleartextCredentials` structs,
depending on the mode set by the value of `EnvelopeMode`:

```
enum {
  base(1),
  customIdentifier(2),
  (255)
} EnvelopeMode;
```

The `base` mode defines `SecretCredentials` and `CleartextCredentials` as follows:

```
struct {
  opaque skU<1..2^16-1>;
} SecretCredentials;

struct {
  opaque pkS<1..2^16-1>;
} CleartextCredentials;
```

The `customIdentifier` mode defines `SecretCredentials` and `CleartextCredentials` as follows:

```
struct {
  opaque skU<1..2^16-1>;
} SecretCredentials;

struct {
  opaque pkS<1..2^16-1>;
  opaque idU<0..2^16-1>;
  opaque idS<0..2^16-1>;
} CleartextCredentials;
```

These credentials are embedded into the following `Envelope` structure with
encryption and authentication.

```
struct {
  InnerEnvelopeMode mode;
  opaque nonce[32];
  opaque ct<1..2^16-1>;
} InnerEnvelope;

struct {
  InnerEnvelope contents;
  opaque auth_tag[Nh];
} Envelope;
```

Also updated the `FinalizeRequest` and `RecoverCredentials` procedures accordingly. Also updated the fixed string to read `OPAQUE01` instead of `OPAQUE00`.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -386,15 +386,12 @@ We first define the core OPAQUE protocol based on a generic OPRF, hash, and MHF 
 including: HMQV, 3DH, and SIGMA-I. {{I-D.sullivan-tls-opaque}} discusses integration with
 TLS 1.3 {{RFC8446}}.
 
-## Data types {#data-types}
+## Credential types and envelope construction {#data-types}
 
-OPAQUE makes use of a structure `Credentials` to store user (client) credentials.
-A `Credentials` structure consists of secret and cleartext `CredentialExtension`
-values. Each `CredentialExtension` indicates the type of extension and carries
-the raw bytes. This specification includes extensions for OPAQUE, including:
+OPAQUE makes use of a structure `Envelope` to store client credentials.
+The `Envelope` structure embeds the following types of credentials:
 
 - skU: The encoded user private key for the AKE protocol.
-- pkU: The encoded user public key for the AKE protocol.
 - pkS: The encoded server public key for the AKE protocol.
 - idU: The user identity. This is an application-specific value, e.g., an e-mail
   address or normal account name.
@@ -405,54 +402,51 @@ Each public and private key value is an opaque byte string, specific to the AKE
 protocol in which OPAQUE is instantiated. For example, if used as raw public keys
 for TLS 1.3 {{?RFC8446}}, they may be RSA or ECDSA keys as per {{?RFC7250}}.
 
-The full `Credentials` encoding is as follows, described using TLS notation
-(see {{RFC8446}}, Section 3).
+These credentials are incorporated in the `SecretCredentials` and `CleartextCredentials` structs,
+depending on the mode set by the value of `EnvelopeMode`:
 
 ~~~
 enum {
-  skU(1),
-  pkU(2),
-  pkS(3),
-  idU(4),
-  idS(5),
+  base(1),
+  customIdentifier(2),
   (255)
-} CredentialType;
-
-struct {
-  CredentialType type;
-  opaque data<0..2^16-1>;
-} CredentialExtension;
-
-struct {
-  CredentialExtension secret_credentials<1..2^16-1>;
-  CredentialExtension cleartext_credentials<0..2^16-1>;
-} Credentials;
+} EnvelopeMode;
 ~~~
 
-secret_credentials
-: OPAQUE credentials which require secrecy and authentication.
-
-cleartext_credentials
-: OPAQUE credentials which require authentication but not secrecy.
-
-Applications MUST include `skU` in `secret_credentials` and `pkS` in either `cleartext_credentials`
-or `secret_credentials`. All other CredentialExtension values are optional. It is RECOMMENDED
-that applications include `pkS` and `idS` in `cleartext_credentials`, as this allows servers
-to not store redundant encryptions of these values for each user in case the server uses the
-same values for multiple users.
-
-Additionally, we assume helper functions `SerializeExtensions` and `DeserializeExtensions`
-which translate a list of `CredentialExtension` structures to and from a unique byte string
-encoding.
-
-OPAQUE uses an `Envelope` structure to encapsulate an encrypted `Credentials` structure.
-It is encoded as follows.
+The `base` mode defines `SecretCredentials` and `CleartextCredentials` as follows:
 
 ~~~
 struct {
+  opaque skU<1..2^16-1>;
+} SecretCredentials;
+
+struct {
+  opaque pkS<1..2^16-1>;
+} CleartextCredentials;
+~~~
+
+The `customIdentifier` mode defines `SecretCredentials` and `CleartextCredentials` as follows:
+
+~~~
+struct {
+  opaque skU<1..2^16-1>;
+} SecretCredentials;
+
+struct {
+  opaque pkS<1..2^16-1>;
+  opaque idU<0..2^16-1>;
+  opaque idS<0..2^16-1>;
+} CleartextCredentials;
+~~~
+
+These credentials are embedded into the following `Envelope` structure with
+encryption and authentication.
+
+~~~
+struct {
+  InnerEnvelopeMode mode;
   opaque nonce[32];
   opaque ct<1..2^16-1>;
-  opaque auth_data<0..2^16-1>;
 } InnerEnvelope;
 
 struct {
@@ -465,13 +459,16 @@ nonce
 : A unique 32-byte nonce used to protect this Envelope.
 
 ct
-: Encoding of encrypted and authenticated credential extensions list (`secret_credentials`).
-
-auth_data
-: Encoding of an authenticated credential extensions list (`cleartext_credentials`).
+: Encoding of encrypted and authenticated `SecretCredentials`.
 
 auth_tag
-: Authentication tag protecting the contents of the envelope.
+: Authentication tag protecting the contents of the envelope,
+covering `InnerEnvelope` and `CleartextCredentials`.
+
+The full procedure for constructing `Envelope` and `InnerEnvelope` from
+`SecretCredentials` and `CleartextCredentials` is described in {{finalize-request}}.
+
+The `EnvelopeMode` value is specified as part of the configuration (see {{configurations}}).
 
 ## Offline registration stage {#offline-phase}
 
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -465,13 +459,14 @@ nonce
...
-: Encoding of encrypted and authenticated credential extensions list (`secret_credentials`).
-
-auth_data
-: Encoding of an authenticated credential extensions list (`cleartext_credentials`).
+: Encoding of encrypted and authenticated `SecretCredentials`.
 
 auth_tag
-: Authentication tag protecting the contents of the envelope.
+: Authentication tag protecting the contents of the envelope,
+incorporating `InnerEnvelope` and `CleartextCredentials`.
```
chris-wood, line 466:
```suggestion
covering `InnerEnvelope` and `CleartextCredentials`.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1038,8 +1012,8 @@ info = "HMQV keys" || I2OSP(len(nonceU), 2) || nonceU
 ~~~
 
 Here, idU and idS are by default set to be equal to the idU and idS supplied as a
-`CredentialExtension` for the envelope; however, if no such extension were supplied,
-then these values are defaulted to pkU and pkS instead.
+credential for the envelope; however, if no such credential was supplied,
+then these values are (indepedently) defaulted to pkU and pkS instead.
```
chris-wood, line 1016:
```suggestion
then these values (independently) default to pkU and pkS instead.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -465,13 +459,14 @@ nonce
...
-: Encoding of an authenticated credential extensions list (`cleartext_credentials`).
+: Encoding of encrypted and authenticated `SecretCredentials`.
 
 auth_tag
-: Authentication tag protecting the contents of the envelope.
+: Authentication tag protecting the contents of the envelope,
+incorporating `InnerEnvelope` and `CleartextCredentials`.
+
+The full procedure for constructing `Envelope` and `InnerEnvelope` from
+`SecretCredentials` and `CleartextCredentials` is described in {{finalize-request}}.
```
chris-wood, line 469:
```suggestion
`SecretCredentials` and `CleartextCredentials` is described in {{finalize-request}}.

The `EnvelopeMode` value is specified as part of a configuration. 
```

chris-wood (commented):
I like this :) I have a couple questions that I hope we can resolve moving forward:

- The mode is currently implicit in registration and login. I assume the expectation is that configurations specify the mode so that it's known to both peers. We should probably say that somewhere. I left a suggestion to include this note, but we probably ought to expand upon it in the configuration section.
- The `RegistrationResponse` definition still includes a list of secret and cleartext credential types. This can be removed given that these are dependent on the mode. 
- Credential construction is somewhat vague in the current description, especially since idU and idS are not passed as parameters to the main registration and login routines. As an implementor, is the expectation that idS comes from the client's wrapper protocol, or is it sent from server to client as part of the registration response? (My understanding is the latter.) Since there are only two modes, we might, for example, define two mode-specific `FinalizeRequest` variants, each of which takes the appropriate parameters. (The base mode variant includes what is currently there now, whereas the custom-mode variant might also include idU (and idS depending on the answer to the previous question).)

Obviously, we can work on these things incrementally and merge in pieces, but I wanted to flag these questions to start.

kevinlewi:
* I added a note in the configurations section: "The EnvelopeMode value is defined in {{data-types}} and must also be agreed upon between the client and server."
* Thanks, good catch. Removed!
* The expectation is that idS should come from the wrapper protocol. I think it is unnecessary to include idS in `RegistrationResponse`, regardless of the mode. Therefore, I think it is unnecessary to change anything here regarding this point. What do you think?

draft-irtf-cfrg-opaque.md
```diff
@@ -1174,6 +1148,9 @@ The OPAQUE AKE protocols are those which are specified in {{instantiations}}.
 Future specifications (such as {{I-D.sullivan-tls-opaque}}) MAY introduce other
 AKE instantiations.
 
+The EnvelopeMode value is defined in {{data-types}} and must also be agreed upon
```
chris-wood, line 1151: This implies (to me) some sort of negotiation. Since the EnvelopeMode is already in the tuple above, I'd drop that bit:

```suggestion
The EnvelopeMode value is defined in {{data-types}}.
```

chris-wood:
> The expectation is that idS should come from the wrapper protocol. I think it is unnecessary to include idS in RegistrationResponse, regardless of the mode. Therefore, I think it is unnecessary to change anything here regarding this point. What do you think?

I think that's fine, but I would make things more explicit. If the wrapper protocol provides them, then it might as well also supply them to the relevant procedures for constructing the envelope. Then we could replace this:

"Create cleartext_credentials with credentials matching those needed to construct the `CleartextCredentials` structure"

with something like:

"Create `CleartextCredentials` structure cleartext_credentials where cleartext_credentials.idU = idU, cleartext_credentials.idS = idS, cleartext_credentials.pkS = pkS."

draft-irtf-cfrg-opaque.md
```diff
@@ -624,45 +612,31 @@ Output:
...
 10. export_key = HKDF-Expand(rwdU, concat(nonce, "ExportKey"), Nh)
 11. ct = xor(pt, pseudorandom_pad)
-12. auth_data = SerializeExtensions(cleartext_credentials)
-13. Create InnerEnvelope contents with (nonce, ct, auth_data)
-14. t = HMAC(auth_key, contents)
+12. auth_data = Serialize(cleartext_credentials)
+13. Create InnerEnvelope contents with (nonce, ct)
+14. t = HMAC(auth_key, concat(contents, auth_data))
 15. Create Envelope envU with (contents, t)
 16. Create RegistrationUpload upload with envelope value (envU, pkU)
```
chris-wood, line 631: Unrelated: should we lift `pkU` to the wrapper protocol, too?

draft-irtf-cfrg-opaque.md
```diff
@@ -598,11 +586,11 @@ Steps:
...
-     (Z, pkS, secret_credentials_list, cleartext_credentials_list)
+     (Z, pkS)
 4. Output (response, kU)
 ~~~
 
-#### FinalizeRequest
+#### FinalizeRequest {#finalize-request}
 
 ~~~
 FinalizeRequest(pwdU, skU, blind, request, response)
```
chris-wood, line 596: Also unrelated, but we can remove `request` here, as it's not used.

chris-wood:
Following up on the comment above, we might simplify things by delegating the "credential construction" logic to the wrapper protocol, which (a) knows the mode and (b) has the relevant information (skU, idU, pkS, idS, etc). That would make the interface to `FinalizeRequest` something like so:

```
FinalizeRequest(pwdU, creds, blind, response)
```

Where `creds` contains the `secret_credentials` and `cleartext_credentials`, already in wire format form, to use when creating the envelope.

bytemare:
Since idu is used on both sides for the AKE info HKDF parameter, shouldn't it always be conveyed to the client?

draft-irtf-cfrg-opaque.md
```diff
@@ -624,45 +612,31 @@ Output:
...
-4. Create secret_credentials with CredentialExtensions matching that
-   contained in response.secret_credentials_list
-5. Create cleartext_credentials with CredentialExtensions matching that
-   contained in response.cleartext_credentials_list
-6. pt = SerializeExtensions(secret_credentials)
+4. Create secret_credentials with credentials matching those needed to
+   construct the `SecretCredentials` structure
+5. Create cleartext_credentials with credentials matching those needed to
+   construct the `CleartextCredentials` structure
+6. pt = Serialize(secret_credentials)
```
bytemare, line 621: If SecretCredentials is always equal to the client secret, wouldn't it be more straightforward to put `pt = Serialize(secret)` ?

draft-irtf-cfrg-opaque.md
```diff
@@ -624,45 +612,31 @@ Output:
...
 7. nonce = random(32)
 8. pseudorandom_pad = HKDF-Expand(rwdU, concat(nonce, "Pad"), len(pt))
 9. auth_key = HKDF-Expand(rwdU, concat(nonce, "AuthKey"), Nh)
 10. export_key = HKDF-Expand(rwdU, concat(nonce, "ExportKey"), Nh)
 11. ct = xor(pt, pseudorandom_pad)
-12. auth_data = SerializeExtensions(cleartext_credentials)
-13. Create InnerEnvelope contents with (nonce, ct, auth_data)
-14. t = HMAC(auth_key, contents)
+12. auth_data = Serialize(cleartext_credentials)
+13. Create InnerEnvelope contents with (nonce, ct)
```
bytemare, line 628: If I'm not mistaken, the InnerEnvelope should also specify the InnerEnvelopeMode, doesn't it?

draft-irtf-cfrg-opaque.md
kevinlewi: It is used in line 468:

```
The full procedure for constructing `Envelope` and `InnerEnvelope` from
`SecretCredentials` and `CleartextCredentials` is described in {{finalize-request}}.
```

draft-irtf-cfrg-opaque.md
kevinlewi: As discussed offline, let's address this API change separately.

draft-irtf-cfrg-opaque.md
kevinlewi: I think it may be beneficial to keep this level of generality here in case we add support for other modes that set SecretCredentials to something else in the future. This will hopefully also coincide with the changes that will be made in #105 which will pass in a generic credentials parameter.

draft-irtf-cfrg-opaque.md
kevinlewi: Good catch, thank you! Added to parameters.

kevinlewi:
Addressed comments, ready for a re-review

draft-irtf-cfrg-opaque.md
chris-wood: I'll address this in a separate PR, since it seems related to the module boundary we discussed.

chris-wood (approved):

Issue #63, opened 2020-10-05
Extraneous parameter pkS in CredentialResponse


kevinlewi:
Why is it necessary to [put pkS in the CredentialResponse struct](https://github.com/cfrg/draft-irtf-cfrg-opaque/blame/master/draft-irtf-cfrg-opaque.md#L788)?

According to [this](https://github.com/cfrg/draft-irtf-cfrg-opaque/blame/master/draft-irtf-cfrg-opaque.md#L460-L464), pkS must be included within the envelope already. So, when the client parses the CredentialResponse struct, shouldn't it be able to obtain pkS from the envelope already?

chris-wood:
That's a good question! I *think* it can probably be removed, but let's see what @hugokraw thinks?

hugokraw:
I did not notice pkS is sent outside the envelope. Not only it would be redundant but pkS is the most important element to be authenticated as part of the envelope. Sending it outside may lead people to accept it without authentication.

chris-wood:
Let's drop it. @kevinlewi, can you please prep a PR to do this?

kevinlewi:
In light of the changes made in #102, this will need to be re-introduced again.

PR #106, opened 2021-01-05
Separate the core and wrapper protocol

### LINKED ISSUE #105: Update wrapper protocol APIs to pass in credentials to FinalizeRequest
Following up on the comment above, we might simplify things by delegating the "credential construction" logic to the wrapper protocol, which (a) knows the mode and (b) has the relevant information (skU, idU, pkS, idS, etc). That would make the interface to `FinalizeRequest` something like so:

```
FinalizeRequest(pwdU, creds, blind, response)
```

Where `creds` contains the `secret_credentials` and `cleartext_credentials`, already in wire format form, to use when creating the envelope.

_Originally posted by @chris-wood in https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/102#issuecomment-749720379_

chris-wood:
Lift wrapper-specific pieces, such as credential information and key material, out of the core protocol, where they are not used. Update the reference implementation to match.

I might ignore the reference implementation for the time being, since it's kind of sloppy. But if you wanted to check it for consistency against the core module API boundary, that would be useful. 

Closes #105.


draft-irtf-cfrg-opaque.md
```diff
@@ -654,12 +657,10 @@ is that which is associated with the OPAQUE configuration (see {{configurations}
...
-#### StoreUserRecord
-
-The StoreUserRecord function stores the tuple `(envU, pkS, skS, pkU, kU)`,
-where envU and pkU are obtained from the input RegistrationUpload message in
-a record associated with the user's account idU. If `skS` and `pkS` are used for
-multiple users, the server can store these values separately and omit them from
+Upon completion of this function, the client MUST send its public key `pkU`, optional
+identity `idU` (depending on the mode), and corresponding envelope `envU` to the server,
+where they are then store as a tuple `(envU, pkU, kU, pkS, skS)`. If `skS` and `pkS` are
+used for multiple users, the server can store these values separately and omit them from
```
kevinlewi, line 663: Why doesn't the client just send an instance of the `RegistrationUpload` function, which includes `pkU` and `envU`? It doesn't need to send `idU` since that is taken care of in the application layer and not necessary for the OPAQUE protocol

draft-irtf-cfrg-opaque.md
```diff
@@ -514,12 +525,11 @@ Once complete, the registration process proceeds as follows:
                                    response
                               <-----------------
 
- record = FinalizeRequest(pwdU, skU, blind, request, response)
+ envU, export_key = FinalizeRequest(pwdU, creds, blind, response)
 
-                                    record
+                                     envU
```
kevinlewi, line 530: I think this should be an instance of the `RegistrationUpload` structure, not `envU`

draft-irtf-cfrg-opaque.md
```diff
@@ -754,40 +751,38 @@ Steps:
...
-2. Create CredentialResponse response with (Z, pkS, envU)
+2. Create CredentialResponse response with (Z, envU)
 3. Output response
 ~~~
 
 #### RecoverCredentials(pwdU, blind, request, response)
 
 ~~~
-RecoverCredentials(pwdU, blind, request, response)
+RecoverCredentials(pwdU, cleartext_creds, blind, response)
```
kevinlewi, line 776: Line 806 should output an instance of `Credentials`, then, right? (Currently it outputs a secret_credentials)

draft-irtf-cfrg-opaque.md
```diff
@@ -453,6 +453,17 @@ struct {
...
 
+The secret and cleartext credentials, along with the mode, are wrapped up in a
+`Credentials` structure, defined as follows:
+
+~~~
+struct {
+  InnerEnvelopeMode mode;
+  SecretCredentials secret_credentials;
+  CleartextCredentials cleartext_credentials;
+} Credentials;
```
kevinlewi, line 464: I believe this will be the only instance of a struct that we define in this document which never gets serialized. Why not just replace all instances with a tuple of (mode, secret_credentials, cleartext_credentials)? And note that the RecoverCredentials function doesn't actually need to return the cleartext_credentials or the mode...

chris-wood: You could, yeah. This is just a helper type to wrap up everything in that tuple.

draft-irtf-cfrg-opaque.md
chris-wood: Hmm, it is nice if it just returns a `Credentials` type. I'll make the change!

draft-irtf-cfrg-opaque.md
chris-wood: I intentionally reverted this so that the wrapper could construct `RegistrationUpload`, but on reflection that's probably not best. I'll revert this bit.

draft-irtf-cfrg-opaque.md
```diff
@@ -720,19 +718,15 @@ data
...
+    Envelope envU;
 } CredentialResponse;
 ~~~
 
 data
 : A serialized OPRF group element.
 
-pkS
-: An encoded public key that will be used for the online authenticated key exchange stage.
-
```
kevinlewi, line 733: pkS should not be removed -- see my last comment on #63 . The problem is that the pkS is no longer included in the envelope.

kevinlewi (approved):
Sorry, I think I missed this with my first pass, but pkS needs to be a part of CredentialResponse (so I would add back pkS everywhere). Other than that, looks good!

kevinlewi (approved):
Looks good!

PR #107, opened 2021-01-06
Pin envelope mode to base or customIdentifier.

chris-wood:
I chose to leave the enumeration and restrict its values to in the text. 

As discussed, this also makes pkU/pkS mandatory in the core OPAQUE messages.


kevinlewi (approved):
Great!

PR #114, opened 2021-01-13
Adding serialization format for credential file

### LINKED ISSUE #89: rename StoreUserRecord to indicating the finalization instead of optional persistence
Just as we dropped the ProtocolMessages part, for the protocol it might be not necessary to call the function StoreUserRecord. For instance my implementation does not care about any kind of persistence or transport of the outcome of the protocol. It might be that the use-case does not need the storage of the user record, for example by just keeping it in memory, or the record might be further processed before storage. Thus i propose to rename the function to `FinalizeUserRecord` instead.

kevinlewi:
Defines the serialization for the "credential file" (the object that is stored server-side and acts as the user's password file).

Closes #89.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -375,6 +375,10 @@ generates an OPRF private and public key. OPAQUE only requires an OPRF private k
 We write `(kU, _) = KeyGen()` to denote use of this function for generating secret key `kU`
 (and discarding the corresponding public key).
 
+We use the types `SerializedElement` and `SerializedScalar` as defined in {{I-D.irtf-cfrg-voprf}},
+along with serialization functions `SerializeElement`, `DeserializeElement`, `SerializeScalar`, and
+`DeserializeScalar`.
+
 # Core Protocol {#protocol}
 
 OPAQUE consists of two stages: registration and authenticated key exchange.
@@ -654,9 +658,19 @@ is that which is associated with the OPAQUE configuration (see {{configurations}
 See {{export-usage}} for details about the output export_key usage.
 
 Upon completion of this function, the client MUST send `record` to the server.
-The server then stores the tuple `(envU, pkU, kU, pkS, skS)`, where `envU` and `pkU`
-are obtained from `record`. If `skS` and `pkS` are used for multiple users, the
-server can store these values separately and omit them from the user's record.
+
+#### Credential File {#credential-file}
+
+The server then constructs and stores the `CredentialFile` object, where `envU` and `pkU`
+are obtained from `record`, and `kU` is retained from the output of `CreateRegistrationResponse`.
+
+~~~
+struct {
+    SerializedScalar kU;
+    opaque pkU<1..2^16-1>;
+    Envelope envU;
+} CredentialFile;
+~~~
 
 ## Online authenticated key exchange stage {#online-phase}
 
@@ -755,20 +769,19 @@ Steps:
 #### CreateCredentialResponse
 
 ~~~
-CreateCredentialResponse(request, pkS, kU, envU)
+CreateCredentialResponse(request, pkS, credentialFile)
 
 Input:
 - request, a CredentialRequest structure
 - pkS, the public key of the server
-- kU, the OPRF key associated with idU
-- envU, the Envelope associated with idU
+- credentialFile, the server's output from registration (see {{credential-file}})
 
 Output:
 - response, a CredentialResponse structure
 
 Steps:
-1. Z = Evaluate(kU, request.data)
-2. Create CredentialResponse response with (Z, pkS, envU)
+1. Z = Evaluate(DeserializeScalar(credentialFile.kU), request.data)
+2. Create CredentialResponse response with (Z, pkS, credentialFile.envU)
 3. Output response
 ~~~
 


draft-irtf-cfrg-opaque.md
```diff
@@ -654,9 +656,19 @@ is that which is associated with the OPAQUE configuration (see {{configurations}
...
-server can store these values separately and omit them from the user's record.
+
+#### Credential File {#credential-file}
+
+The server then constructs and stores the `CredentialFile` object, where `envU` and `pkU`
+are obtained from `record`, and `kU` is retained from the output of `CreateRegistrationResponse`.
+
+~~~
+struct {
+    Scalar kU<1..2^16-1>;
```
chris-wood, line 667:
```suggestion
    SerializedScalar kU;
```

chris-wood: I assume we want this to be serialized for storage, right?

draft-irtf-cfrg-opaque.md
```diff
@@ -755,20 +767,19 @@ Steps:
...
-- envU, the Envelope associated with idU
+- credentialFile, the server's output from registration (see {{credential-file}})
 
 Output:
 - response, a CredentialResponse structure
 
 Steps:
-1. Z = Evaluate(kU, request.data)
-2. Create CredentialResponse response with (Z, pkS, envU)
+1. Z = Evaluate(credentialFile.kU, request.data)
```
chris-wood, line 781: Assuming you take the suggestion above, I'd then deserialize `kU` before use.

chris-wood (approved):
Since `CredentialFile` is meant to be opaque storage handled by OPAQUE, I think everything in the envelope ought to be serialized. Since the serialized format of a `Scalar` depends on the OPRF group, this does assume that the OPAQUE configuration (and OPRF suite) is known. What prevents a `CredentialFile` from one configuration being used in another configuration?

draft-irtf-cfrg-opaque.md
kevinlewi: I could not find "SerializedScalar" in the VOPRF doc. Is there a definition for this somewhere?

kevinlewi:
I think we should leave it up to a wrapper implementation to add a version around the bytes of the `CredentialFile`, no?

kevinlewi:
Comments addressed -- ready to merge

bytemare:
I'm not sure I have a specific question here, but may I share my thoughts on the subject?

I have been playing with different ideas on my side, and found that due to the different variable parameters at play it is a quite complex subject to optimize.

- **Envelope Mode**
If the envelope is serialized and opaque, the server still needs a way to know the EnvelopeMode to communicate the correct "Credentials" structure to the client.

- **OPRF key**
As @chris-wood pointed out, the encoding of an OPRF key is tied to the suite and is not encoded into the key itself. Hence, the most straightforward thing to do is to have another field specifying the suite the key belongs to. It's a single byte, and can't hurt that much, I believe.
If one wants to separate kU from the envelope and store it elsewhere, keys belonging to the same group could be stored in the same table. Thus, the user record only references the kU in another table and doesn't care about the suite.

- **AKE keys**
Given a complex environment with different conditions that open the possibility that different AKEs are used, the user record might need to keep track of that as well. Either a reference to the key pair and their type (group and AKE) or stored in a "table" dedicated to that key pair.

- **MHF Params**
The key derivation functions have different impacts on different platforms and may impact the user experience if the login takes too long due to unsuited parameters. From what I understand, this hasn't been thoroughly discussed, yet (and rightfully so). But if different configurations were to be used for different clients, they have to be saved with the user record.


A "naive" version of my implementation sees the complete information tied to a user record grow quite rapidly:
```
UserRecord {
    HumanUserID  // Human-memorizable, modifiable user identifier
    UUID               // Unique long-term user identifier, or idu
    UserPublicKey
    EnvelopeMode 
    Envelope     
    ServerID     
    OprfSecret   
    ServerAkeID  
    Parameters {
        Ciphersuite
        Hash
        AKE
        MHFParams {
            AlgoID
            time, memory, threads
            KeyLength
        }
    }
}

ServerAke {
    ID             // unique ID for this key pair
    Ake          // the AKE protocol they are used in
    Group    
    Hash     
    SecretKey
    PublicKey
}
```

stef:
i'm not sure this is a good idea, different use-cases might not need all the bells and whistles. i think i asked in an issue if this is about the standardization of the protocol or it goes beyond that and standardizes also serialization of all kind of wire and storage elements. in implementations using only one instantiation of the protocol the parameters are fixed, and don't need to be stored. if an implementation does not follow the spec and does not store this redundant info in that case, will it still be standards compliant? also some implementations might want to go beyond the currently specified two envelope setups and may want to make all credential extensions configurable, i can see reasons for all configurations. i think there should be a limit of what is part of the standard, and what is free for implementations to decide themselves, and again depending on different configurations of how credential extensions are created this can lead to very different storage formats. my take is to not specify neither the storage format of the record, nor actually the configuration of the envelope. test vectors can still be provided for the protocol, especially if serialization is not fixed, in that case the test vectors can be provided as results of the various steps of the protocol, there is no combinatoric explosion.

stef:
to quote Antoine de Saint-Exupery:
> Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away .

stef:
imagine an implementation that stores (parts of) the record in a smartcard, a cortex m[034], a cars ECU, or a yubikey. an implementation that has one pkS and one that has separate pkS for each user. it makes sense to have different implementations to be interoperable with each other on the protocol level, a client and a server running different implementations. but the records, do not need to be specified.

chris-wood:
The standard needs to specify ways to interpret envelopes for the sake of security and interoperability. (What's included in the envelope is critical to the overall security of OPAQUE.) 

The `mode` parameter was introduced precisely to allow the sort of extensibility you're after. For example, if neither of the envelope formats work for your use case, you may pick a new mode and its corresponding envelope structure. The existing modes were chosen because we believe they represent the minimal information needed for most applications.

Regarding the CredentialFile format, this change describes the structure and format of what minimally needs to be *associated* with a given user's record. It does not mandate how server implementations store this information.

bytemare:
Hello Stefan,

my point wasn't to say "hey, this is how I think we should do it, let's put everything in there to have a perfect thingy" :)
As I said, this is the naive structure for my proof-of-concept, and, believe me, I don't want to keep it like this.

I'm aligned with you, and I think we all agree. I believe that there is a _sweet spot_ that OPAQUE shall reach, a balance between the core necessities and minimize implementation pitfalls that might introduce vulnerabilities in the usage of OPAQUE.

stef:
> The standard needs to specify ways to interpret envelopes for the sake of security and interoperability. (What's included in the envelope is critical to the overall security of OPAQUE.)

agreed. security and interop are the goals of an opaque standard.

> The `mode` parameter was introduced precisely to allow the sort of extensibility you're after.
> For example, if neither of the envelope formats work for your use case, you may pick a new mode and its corresponding envelope structure. 

oh? i guess i missed the part where it says that this is an optional extensible parameter, and not necessary to be conformant with the spec. i guess i suffer from patch reading fatigue trying to keep up with all the changes, sorry for that.

> The existing modes were chosen because we believe they represent the minimal information needed for most applications.

is there any documentation on how this belief was established? for example why not chose the minimal envelope, with 0 credentialextensions as a base mode isn't that the minimal information already needed for most applications?
 
> Regarding the CredentialFile format, this change describes the structure and format of what minimally needs to be _associated_ with a given user's record. It does not mandate how server implementations store this information.

maybe calling it a file is misguiding then, perhaps it might make sense to change it to something like:
The following values need to be stored by the server indexed by the idU: kU, envU, pkU.
i do disagree with the part: 
> describes the structure and format
what if i want to store kU in a secure element, or sgx (trying not to laugh hard...) while the rest can be stored outside of it? or alternatively, if there is cleartext elements in the envelope that can be deduped in some way, why store them in the envelope, pkS and idS are some simple candidates to have not stored in the envU, but only reassembled before sending of the CreateCredentialResponse response.

chris-wood:
> is there any documentation on how this belief was established? for example why not chose the minimal envelope, with 0 credentialextensions as a base mode isn't that the minimal information already needed for most applications?

There are no more extensions, so I think this is already addressed :) Please have a look at the latest version of the draft here. (Sorry, many things are changing, and I understand it's hard to follow!)

> maybe calling it a file is misguiding then, perhaps it might make sense to change it to something like:

We have a structure for applications which do not want to store things in some special way, i.e., they want to write the "file" as presented directly to storage.

> what if i want to store kU in a secure element, or sgx (trying not to laugh hard...) while the rest can be stored outside of it? or alternatively, if there is cleartext elements in the envelope that can be deduped in some way, why store them in the envelope, pkS and idS are some simple candidates to have not stored in the envU, but only reassembled before sending of the CreateCredentialResponse response.

As above, the file format *does not* preclude implementation- or application-specific storage for these things. I recommend having a look at the latest version of the spec to see how things are described. 

Alternatively, if it would help, I'd be happy to hop on a call to discuss these issues in more depth. :-)

PR #125, opened 2021-01-26
Structural refactor

chris-wood:
This is a pretty invasive change that tries to better organize the document and protocol description into its two steps: registration and authentication. It also includes more wire format details for the 3DH instantiation, and removes HMQV and SIGMA-I in the process. (We can bring them back in a future change as needed.) The outline now looks like so:

```
   1.  Introduction  . . . . . . . . . . . . . . . . . . . . . . . .   3
     1.1.  Requirements Notation . . . . . . . . . . . . . . . . . .   5
     1.2.  Notation  . . . . . . . . . . . . . . . . . . . . . . . .   5
   2.  Cryptographic Protocol and Algorithm Dependencies . . . . . .   6
   3.  Offline Registration  . . . . . . . . . . . . . . . . . . . .   7
     3.1.  Credential Storage  . . . . . . . . . . . . . . . . . . .   8
     3.2.  Registration Messages . . . . . . . . . . . . . . . . . .  10
     3.3.  Registration Functions  . . . . . . . . . . . . . . . . .  11
       3.3.1.  CreateRegistrationRequest . . . . . . . . . . . . . .  11
       3.3.2.  CreateRegistrationResponse  . . . . . . . . . . . . .  11
       3.3.3.  FinalizeRequest . . . . . . . . . . . . . . . . . . .  11
       3.3.4.  CredentialFile  . . . . . . . . . . . . . . . . . . .  13
   4.  Online Authenticated Key Exchange . . . . . . . . . . . . . .  13
     4.1.  Credential Retrieval  . . . . . . . . . . . . . . . . . .  13
       4.1.1.  Credential Retrieval Messages . . . . . . . . . . . .  14
       4.1.2.  Credential Retrieval Functions  . . . . . . . . . . .  14
     4.2.  AKE Instantiations  . . . . . . . . . . . . . . . . . . .  16
       4.2.1.  Key Schedule Utility Functions  . . . . . . . . . . .  17
       4.2.2.  OPAQUE-3DH Instantiation  . . . . . . . . . . . . . .  18
       4.2.3.  Alternate AKE instantiations  . . . . . . . . . . . .  20
   5.  Configurations  . . . . . . . . . . . . . . . . . . . . . . .  20
```

cc @hugokraw, @crockeea

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -299,6 +299,12 @@ and SIGMA {{SIGMA}}. In general, the modularity of OPAQUE's design makes it
 easy to integrate with additional AKE protocols, e.g., IKEv2, and with future
 ones such as those based on post-quantum techniques.
 
+OPAQUE consists of two stages: registration and authenticated key exchange.
+In the first stage, a client registers its password with the server and stores
+its encrypted credentials on the server. In the second stage, a client obtains
+those credentials, recovers them using the user's password, and subsequently uses
+them as input to an AKE protocol.
+
 Currently, the most widely deployed PKI-free aPAKE is SRP {{?RFC2945}}, which is
 vulnerable to pre-computation attacks, lacks a proof of security, and is less efficient
 relative to OPAQUE. Moreover, SRP requires a ring as it mixes addition and
@@ -384,26 +390,49 @@ OPAQUE relies on the following protocols and primitives:
 Note that we only need the base mode variant (as opposed to the verifiable mode
 variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}.
 
-# Core Protocol {#protocol}
+# Offline Registration {#offline-phase}
 
-OPAQUE consists of two stages: registration and authenticated key exchange.
-In the first stage, a client registers its password with the server and stores
-its encrypted credentials on the server. In the second stage, a client obtains
-those credentials, unlocks them using the user's password and subsequently uses
-them as input to an AKE protocol.
+Registration is executed between a user U (running on a client machine) and a
+server S. It is assumed S can identify U and the client can
+authenticate S during this registration phase. This is the only part
+in OPAQUE that requires an authenticated channel, either physical, out-of-band,
+PKI-based, etc. This section describes the registration flow, message encoding,
+and helper functions. Moreover, U has a key pair (skU, pkU) for an AKE protocol
+which is suitable for use with OPAQUE; See {{online-phase}}. (skU, pkU) may be
+randomly generated for the account or provided by the calling client.
+Clients MUST NOT use the same key pair (skU, pkU) for two different accounts.
 
-Both registration and authenticated key exchange stages require running an OPRF protocol.
-The latter stage additionally requires running a mutually authenticated
-key exchange protocol using credentials recovered after the OPRF protocol completes.
-(The key exchange protocol MUST satisfy forward secrecy and the KCI requirement
-discussed in {{security-considerations}}.)
+To begin, U chooses password pwdU, and S chooses its own pair of private-public
+keys skS and pkS for use with the AKE. S can use the same pair of keys with
+multiple users. These steps can happen offline, i.e., before the registration phase.
+Once complete, the registration process proceeds as follows:
 
-We first define the core OPAQUE protocol based on a generic OPRF, hash, and MHF function.
-{{instantiations}} describes specific instantiations of OPAQUE using various AKE protocols,
-including: HMQV, 3DH, and SIGMA-I. {{I-D.sullivan-tls-opaque}} discusses integration with
-TLS 1.3 {{RFC8446}}.
+~~~
+ Client (pwdU, creds)                               Server (skS, pkS)
+ --------------------------------------------------------------------
+ (request, blind) = CreateRegistrationRequest(pwdU)
 
-## Credential types and envelope construction {#data-types}
+                               request
+                      ------------------------->
+
+            (response, kU) = CreateRegistrationResponse(request, pkS)
+
+                               response
+                      <-------------------------
+
+ (record, export_key) = FinalizeRequest(pwdU, creds, blind, response)
+
+                                record
+                      ------------------------->
+~~~
+
+Both client and server MUST validate the other party's public key before use.
+See {{validation}} for more details.
+
+Upon completion, the S stores U's credentials for later use. See {{credential-file}}
+for a recommended storage format.
+
+## Credential Storage {#credential-storage}
 
 OPAQUE makes use of a structure `Envelope` to store client credentials.
 The `Envelope` structure embeds the following types of credentials:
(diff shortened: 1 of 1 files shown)


hugokraw:
This re-structuring looks good, in particular it needs to correct the impression people had that the AKE part is not an integral (and essential) part of the specification.

As for the presentation of HMQV of SIGMA, I want to have a sketch of them in the document as additional examples, especially given how simple it is to do it on the basis of the 3DH description. For HMQV it just means defining the key derivation and for SIGMA it is not much more than that, and it gives a different type of protocol, in particular one that would be used with TLS and IKE.  Since this  is more on the informational side, I would move them to an appendix.

Regarding the sentence 
`any post-quantum (PQ) AKE protocol that has the message flow above`
I am not sure what the "message flow above" refers to but we should not prescribe a particular flow. Definitely not a 3-flow KE since KEM-based AKE would most probably not preserve that.

draft-irtf-cfrg-opaque.md
```diff
@@ -1106,55 +989,48 @@ Likewise, `K3dh` is computed by the server as follows:
...
-as `epkS^eskU` and by servers as `epkU^eskS`.
+It is possible to instantiate OPAQUE with other AKEs, such as HMQV {{HMQV}} and SIGMA-I.
+HMQV is similar to 3DH but varies in its key schedule. SIGMA-I uses digital signatures
+rather than static DH keys for authentication. Specification of these instantiations is
+left to future documents.
+
+OPAQUE may also be instantiated with any post-quantum (PQ) AKE protocol that has the message
+flow above and security properties (KCI resistance and forward secrecy) outlined
+in {{security-considerations}}. This document does not specify such an instantiation.
+Note that such an instantiation is not quantum safe unless the OPRF and data encryption schemes
```
crockeea, line 1023: The data encryption scheme for the envelope encryption is already PQ, and to me, any mention of "data encryption" outside the key exchange step is outside the scope of this doc.

crockeea:
Is there a rendered version of this somewhere?

draft-irtf-cfrg-opaque.md
hugokraw: How about the following formulation (it refers to 'application data' explicitly to avoid confusion of what data we are talking about):
Note that such an instantiation is not quantum safe unless the OPRF is quantum safe. However, an OPAQUE instantiation where the AKE is quantum safe, but the OPRF is not, would still ensure the confidentiality of application data encrypted under session_key (or a key derived from it) with a quantum-safe encryption function.

hugokraw: Btw, we never say what one does with session_key. We may not want to add a derivation of traffic keys as in TLS but we should comment that deriving such keys would be the typical use although applications may have other uses.

draft-irtf-cfrg-opaque.md
```diff
@@ -384,26 +390,49 @@ OPAQUE relies on the following protocols and primitives:
...
+                               response
+                      <-------------------------
+
+ (record, export_key) = FinalizeRequest(pwdU, creds, blind, response)
 
-## Credential types and envelope construction {#data-types}
+                                record
+                      ------------------------->
+~~~
+
```
kevinlewi, line 428: Nit: Perhaps at this point, we should clarify that a bunch of the variables depicted above will be clarified later? Or maybe we can reference a description to them somewhere. After reading this, I become immediately confused... but maybe it is just me :)

draft-irtf-cfrg-opaque.md
```diff
@@ -384,26 +390,49 @@ OPAQUE relies on the following protocols and primitives:
...
+Both client and server MUST validate the other party's public key before use.
+See {{validation}} for more details.
+
+Upon completion, the S stores U's credentials for later use. See {{credential-file}}
```
kevinlewi, line 432: Nit: "the S stores U's credentials" -> "S stores U's credentials" / "the server stores the user's credentials"

draft-irtf-cfrg-opaque.md
```diff
@@ -679,19 +668,29 @@ struct {
...
 the key exchange protocol is 3DH, as described in {{SecHmqv}}, then the private and
 public keys must be Diffie-Hellman keys. At the end, the client proves the user's
 knowledge of the password, and both client and server agree on a mutually authenticated
 shared secret key.
 
-This section describes the message flow, encoding, and helper functions used in this stage.
+OPAQUE produces two outputs: a session secret and an export key. The export key may be used
+for additional application-specific purposes. For example, one might expand the use of OPAQUE
+with a credential-retrieval functionality that is separate from the contents of the `Envelope`
+structure. The exporter_key MUST NOT be used in any way before the HMAC value in the envelope
```
kevinlewi, line 686: Rename exporter_key to export_key

draft-irtf-cfrg-opaque.md
```diff
@@ -934,85 +870,89 @@ HKDF uses Hash as its underlying hash function, which is the same as that
 which is indicated by the OPAQUE instantiation. Note that the Label parameter
 is not a NULL-terminated string.
 
-## Instantiation with HMQV and 3DH {#SecHmqv}
+### OPAQUE-3DH Instantiation {#SecHmqv}
```
kevinlewi, line 873: Nit: Perhaps change the anchor link from `#SecHmqv` to something else?

draft-irtf-cfrg-opaque.md
```diff
@@ -1106,55 +989,48 @@ Likewise, `K3dh` is computed by the server as follows:
...
+Ke2 and then used as input to XOR with the plaintext. In pseudocode, this is
+done as follows:
 
 ~~~
-info = "SIGMA-I keys" || I2OSP(len(nonceU), 2) || nonceU
-                      || I2OSP(len(nonceS), 2) || nonceS
-                      || I2OSP(len(idU), 2) || idU
-                      || I2OSP(len(idS), 2) || idS
+info_pad = HKDF-Expand(Ke2, "encryption pad", len(info2))
+einfo = xor(info_pad, info2)
```
kevinlewi, line 1010: nit: info2 does not seem to be defined anywhere

kevinlewi (approved):
This is fantastic! Left some comments but overall this should help to clear things up a bit.

@crockeea, unfortunately I think you just have to clone the PR and render it locally...

And perhaps we can defer the descriptions of HMQV + SIGMA-I to the Appendix, but also in a later PR since this one already has a bunch of changes.

stef:
@crockeea you can find a rendered version at https://github.com/cfrg/draft-irtf-cfrg-opaque/blob/42c6cfc7fe40015d5e9e4956e0bfe76045bdc357/draft-irtf-cfrg-opaque.md

draft-irtf-cfrg-opaque.md
chris-wood: Yep, good idea!

draft-irtf-cfrg-opaque.md
chris-wood: Thanks, @hugokraw -- I applied your suggested text, and added a note describing that applications can use `session_key` to derive additional keying material.

Issue #121, opened 2021-01-22
Change all occurrences of variable-length public/private keys to be fixed length?


kevinlewi:
In the current version of the draft, there are occurrences such as:

```
struct {
    SerializedElement data;
    opaque pkS<1..2^16-1>;
} RegistrationResponse;
```

```
struct {
    opaque pkU<1..2^16-1>;
    Envelope envU;
} RegistrationUpload;
```

Should we change these to instead say:
```
    opaque pkS[LK];
```

where `LK` is a fixed constant dependent on the group representation?

@chris-wood

chris-wood:
This seems fine, but I would say that `LK` is not a function of the *group*, but simply the fixed length of an AKE public key. @kevinlewi, could you please propose text?

hugokraw:
See my comment on #132

PR #137, opened 2021-02-03
Limiting opaque key types to 255 bytes

### LINKED ISSUE #132: HKDF-Expand limits
RFC 5869 limits the output length of Expand to 255*HashLen, but the envelope ciphertext has a larger capacity. (In theory, HKDF can produce any number of bytes, but we should abide by the interface limitations.)

kevinlewi:
Changes all opaque key types to be at most 255 bytes (instead of 2^16-1). Also changed encrypted_creds in the envelope to have the same restriction.

Note: Still need to update the test vectors to match this new change.

Closes #132

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -465,11 +465,11 @@ The `base` mode defines `SecretCredentials` and `CleartextCredentials` as follow
 
 ~~~
 struct {
-  opaque client_private_key<1..2^16-1>;
+  opaque client_private_key<1..255>;
 } SecretCredentials;
 
 struct {
-  opaque server_public_key<1..2^16-1>;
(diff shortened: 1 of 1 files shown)


hugokraw:
> Changes all opaque key types to be at most 255 bytes (instead of 2^16-1). Also changed encrypted_creds in the envelope to have the same restriction.

This may be too restrictive with PQ algorithms. So if you set that limit, a discussion will be needed about what you do when the algorithm you want to deploy has more than 255 bytes. I will check more about lattice-based schemes and how well they accommodate expanding a PRG seed (namely, what's the computational cost of expanding the seed to a full key).

kevinlewi:
Putting this in draft mode so that we can address #121 with this PR as well.

kevinlewi:
Closing in favor of #141

Issue #69, opened 2020-10-21
Specify harden parameters


chris-wood:
Clients need a way to recover credential information during registration. So we should either (a) fix some Harden parameters, or (b) have servers send these during the registration and authentication flows. 

@kevinlewi, what do you think?

kevinlewi:
Can you clarify what you mean by (a)? If you mean to provide concrete settings, I think we may want to avoid that since these numbers can become out of date, and are application-dependent. For (b), I think we should avoid this as well, unless we start also deciding to send information about group representation / hash functions during these flows. Also, there is no way for the server to enforce that the client follows the harden parameters, and so these are at best recommendations.

My approach would be to allow for parameterizing the protocol based on harden parameters. So in the same way that we can have OPAQUE with Ristretto points versus Edwards, or OPAQUE with SHA3 versus Blake3, the choice of a hardened hash function should just be another parameter.

chris-wood:
>  If you mean to provide concrete settings, I think we may want to avoid that since these numbers can become out of date, and are application-dependent. 

Yeah, that's what I mean, and that's the risk. 

> My approach would be to allow for parameterizing the protocol based on harden parameters. So in the same way that we can have OPAQUE with Ristretto points versus Edwards, or OPAQUE with SHA3 versus Blake3, the choice of a hardened hash function should just be another parameter.

I think there's generally too much flexibility in what's currently in the document. We have ciphersuites, configurations, optional parameters, etc. All of this runs the risk of making implementations challenging to get right. I would be in favor of something a bit more rigid. Could you please draft a PR to take care of this?

kevinlewi:
I might have misinterpreted, but I was not suggesting to modify the current proposal in any way, except to suggest that when producing test vectors, we have a set which instantiates Harden with a specific function.

So my proposal is to keep everything in the draft the same, and when presenting test vectors, we provide annotated options such as:
- OPAQUE-X25519-3DH-SHA256-Scrypt10k,
- OPAQUE-Ristretto255-SigmaI-Blake3-Argon2i100k,
- etc.

This means that we will be supporting the increased flexibility that you may be against.

Btw, while we are on the topic, the flexible parameters I am currently tracking are:
- Group representation
- Keypair format
- AKE protocol
- Hash function
- Harden function

(Also represented in code [here](https://github.com/novifinancial/opaque-ke/blob/master/src/ciphersuite.rs#L15-L23))

Let me know if you think I am missing any others...

[...]

PR #138, opened 2021-02-04
Add back alternate AKE sketches

### LINKED ISSUE #128: HMQV and SIGMA to an Appendix
I want the HMQV and SIGMA specification to appear in an appendix. It serves as illustration and guidance for these and other instantiations.  Note that there is text referring to these protocols (e.g., introduction) that needs to be updated.

chris-wood:
Also, nail down the MHF parameters. We don't currently allow them to be negotiated *in* the protocol. (The wrapper protocol may choose to allow negotiation, but that negotiation would yield a different configuration.)

Closes #128 and #69.

cc @hugokraw


kevinlewi:
Change overall looks good to me. One nit: Perhaps preface the titles of the last few sections with "Appendix: Test Vectors" and "Appendix: Alternate AKE instantiations", just to make it extra clear?

chris-wood:
These are already rendered in the appendix:

```
   Appendix B.  Alternate AKE instantiations
     B.1.  HMQV Instantiation Sketch
     B.2.  SIGMA-I Instantiation Sketch
```

Probably adding "Appendix" to the section headers would be redundant?

kevinlewi:
> These are already rendered in the appendix:
> 
> ```
>    Appendix B.  Alternate AKE instantiations
>      B.1.  HMQV Instantiation Sketch
>      B.2.  SIGMA-I Instantiation Sketch
> ```
> 
> Probably adding "Appendix" to the section headers would be redundant?

Oops, didn't realize that. Great!

kevinlewi:
Test vectors look good!

kevinlewi (approved):

hugokraw:
I see no point in adding SIGMA to the appendix the way it is now. If you put a sketch of it in the appendix then let's put a sketch of it, not just one sentence. The only substantial difference with the description of 3DH and HMQV is the signature. The rest follows the same pattern as 3DH and HMQV.  I believe that this description will be useful and since we do not provide test vectors then it does not have an extra cost.

PR #139, opened 2021-02-04
Match identifiers with main document.

armfazh:
```txt
enum {
  base(1),
  custom_identifier(2),
  (255)
} EnvelopeMode;
```

### DIFF
--- a/poc/opaque_messages.sage
+++ b/poc/opaque_messages.sage
@@ -38,8 +38,8 @@ else:
 #   CleartextCredentials cleartext_credentials;
 # } Credentials;
 
-envelope_mode_base = 0x00
-envelope_mode_custom_identifier = 0x01
+envelope_mode_base = 0x01
+envelope_mode_custom_identifier = 0x02
 
 def deserialize_secret_credentials(data):
     skU, offset = decode_vector(data)


kevinlewi:
Good catch, @armfazh, thanks!! @chris-wood should we instead change base to be 0 and custom_identifier to be 1 in the main document, or go ahead with this change?

chris-wood:
Let's just update the text to start at 0. I'm not sure why we had them start at 1? @armfazh, could you please submit a PR with that change? (I can also update #138 to do that.)

bytemare:
In other designs, IDs start at 1 to allow for a default value or a value that doesn't pin an exact identifier. E.g if the client starts with a configuration set where the mode is 0, the mode is not forced to be one or another, meaning the client can then determine at runtime what it prefers or what's available (given the context the client may or may not have access to identifiers, and will only decide at protocol execution).
Also, the OPRF ciphersuites start at 1 as well :)

chris-wood:
> In other designs, IDs start at 1 to allow for a default value or a value that doesn't pin an exact identifier. E.g if the client starts with a configuration set where the mode is 0, the mode is not forced to be one or another, meaning the client can then determine at runtime what it prefers or what's available (given the context the client may or may not have access to identifiers, and will only decide at protocol execution).

I don't follow this rationale. Modes are part of the configuration -- they're not 'negotiated' in any way.

> Also, the OPRF ciphersuites start at 1 as well :)

Yeah, though it's somewhat arbitrary. I'll just take this change and update the test vectors in #138.

chris-wood:
Applied this change to #138. Thanks, @armfazh!

PR #141, opened 2021-02-10
Adding Npk / Nsk parameter to fix public/private key sizes

### LINKED ISSUE #121: Change all occurrences of variable-length public/private keys to be fixed length?

kevinlewi:
Gets rid of the variable-length encoding of the client/server public/private keys, and restricts them to be of a fixed length (denoted by the parameters `Npk` and `Nsk`). Note that `Npk` and `Nsk` are parameters that must be specified for each AKE instantiation -- this gives the flexibility of future AKE instantiations that are PQ to use a larger `Npk` / `Nsk` if necessary.

Closes #121 and #132

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -383,6 +383,10 @@ OPAQUE relies on the following protocols and primitives:
     hash is determined by the chosen OPRF group.
   - Nh: The output size of the Hash function.
 
+- Authenticated Key Exchange (AKE, {{instantiations}}):
+  - Npk: The size of the public keys used for the key exchange protocol
+  - Nsk: The size of the private keys used for the key exchange protocol
+
 - Memory Hard Function (MHF):
   - Harden(msg, params): Repeatedly apply a memory hard function with parameters
     `params` to strengthen the input `msg` against offline dictionary attacks.
@@ -466,11 +470,11 @@ The `base` mode defines `SecretCredentials` and `CleartextCredentials` as follow
 
 ~~~
 struct {
-  opaque client_private_key<1..2^16-1>;
+  opaque client_private_key[Nsk];
 } SecretCredentials;
 
 struct {
-  opaque server_public_key<1..2^16-1>;
+  opaque server_public_key[Npk];
 } CleartextCredentials;
 ~~~
 
@@ -478,11 +482,11 @@ The `custom_identifier` mode defines `SecretCredentials` and `CleartextCredentia
 
 ~~~
 struct {
-  opaque client_private_key<1..2^16-1>;
+  opaque client_private_key[Nsk];
 } SecretCredentials;
 
 struct {
-  opaque server_public_key<1..2^16-1>;
+  opaque server_public_key[Npk];
   opaque client_identity<0..2^16-1>;
   opaque server_identity<0..2^16-1>;
 } CleartextCredentials;
@@ -495,7 +499,7 @@ encryption and authentication.
 struct {
   EnvelopeMode mode;
   opaque nonce[32];
-  opaque encrypted_creds<1..2^16-1>;
+  opaque encrypted_creds[Nsk];
 } InnerEnvelope;
 
 struct {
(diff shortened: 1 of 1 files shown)


chris-wood (approved):
LGTM with some suggestions -- thanks!

kevinlewi:
- Added Npk/Nsk instead of just Nk.
- Added extra text to the Alternate AKEs section

kevinlewi:
Btw, @chris-wood  -- landing this will require an update to the test vectors. Should that be done in this change, or is it ok to land this first and expect that the test vectors will be updated in a later PR?

chris-wood:
We (I) can do it in a separate PR. Before we merge this, I'd like to hear from @hugokraw about the proposal.

hugokraw (commented):
I am not sure how I feel about fixing the length of keys. I guess it would be defined as part of a ciphersuite (or something that looks like it). Note that different representations of the same curves may have different public key lengths.

This one I don't like:

```
 Future specifications which have keys exceeding this size should specify a
 mechanism by which private keys and their corresponding public keys can be
 deterministically derived from a fixed-length seed.
```

That seems an overkill way of dealing with the HKDF limit of 255*hlen. A simpler mechanism is to expand HKDF-expand, e.g. via concatentation of Expand(prk, info1, L1) and Expand(prk, info2, L2), etc.

chris-wood:
> I am not sure how I feel about fixing the length of keys. I guess it would be defined as part of a ciphersuite (or something that looks like it). Note that different representations of the same curves may have different public key lengths.

Yeah, this is the idea. A configuration (ciphersuite) would fully identify the AKE, including the key(s) used.

> That seems an overkill way of dealing with the HKDF limit of 255*hlen. A simpler mechanism is to expand HKDF-expand, e.g. via concatentation of Expand(prk, info1, L1) and Expand(prk, info2, L2), etc.

While true, I'd prefer not to have HKDF-specific workarounds in the spec. If anything, I'd suggest we factor out HKDF to a generic KDF that does not have an input limitation. Either way, this particular issue seems orthogonal to the overall change here (moving to fixed-length keys). 

I suggest we land this PR, and then use #132 to track a fix to the input limit.

hugokraw:
> I'd prefer not to have HKDF-specific workarounds in the spec.

I was not suggesting to specify that workaround as part of the spec. I was mostly "complaining" about telling people to use what can be an involved procedure (full key generation from a seed) when there are simpler straightforward workarounds for the HKDF length issue.  If you are already suggesting some solutions, I suggest you mention this simpler option. 

Just for fun I was looking at what discussion I had with Pasi Eronen when writing the HKDF draft and I said 

`The other issue that I am not totally happy with is the counter in the calculation of T(i)'s. I agree that for typical IETF applications, namely, protocols where HKDF is used to derive shared keys, the 2^8*2^5 bytes of key material (assuming SHA-256) should suffice. But I see the definition of HKDF as serving other purposes, such as more general pseudo-random generation, where larger amounts of output bits may be required. Should we say that applications may decide to use HKDF with longer counters? Or should we suggest a way of "re-keying" the PRF as I indicated in my previous message to you?`

(the "serving other purposes" is well exemplified here)

Pasi answered:

`About the counter length: I suggest we leave it as-is for now. 
(BTW, an application that needs more than 8 KB could also call
the KDF several times with different "info" values).`

The eternal tension between generality and complexity. Personally, I agree that too much flexibility is not good. But limiting generality also has its price and these things need to be weighted carefully. What is useful is to provide advise on points that may be too limiting (e.g., I wish we had a remark about the hkdf length workaround)

chris-wood:
Yeah, I hear what you're saying :-) I moved this comment over to #132, where we can address this issue in a future-proof way! For now, I'm merging this change.

Issue #142, opened 2021-02-10
KE2 and KE3 transcript mac pre-hashing


bytemare:
In the poc, KE2 and KE3 have their transcript hashed before feeding it into the HMAC

KE2 https://github.com/cfrg/draft-irtf-cfrg-opaque/blob/master/poc/opaque_ake.sage#L160
KE3 https://github.com/cfrg/draft-irtf-cfrg-opaque/blob/master/poc/opaque_ake.sage#L207

But the draft doesn't mention these pre-hashing steps. This changes the HMAC output, thus don't conform with the design in the draft.
cc @chris-wood and @kevinlewi

chris-wood:
Fixed in #147.

PR #144, opened 2021-02-13
Address #142 and minor suggestions to the draft

bytemare:
In _opaque_ake.sage_ drop pre-hashing and carry only the transcript in the state + minor renaming to avoid confusion.

Some typo fixing in the draft and suggestions.


stef:
can you please produce a diff that is not a complete replacement of the changed files. reviewing the changes like this is a bit uneconomic.

bytemare:
Oh, I see. I don't know how this happened. Do you know how I can avoid that?

bytemare:
Thanks, @stef! I submitted the same change in #145, with a correct diff :)

PR #145, opened 2021-02-13
Address #142 and minor suggestions to the draft

bytemare:
In opaque_ake.sage drop pre-hashing and carry only the transcript in the state + minor renaming to avoid confusion.

Some typo fixing in the draft and suggestions for wording.

cc @chris-wood


stef:
why not hash the transcript?

bytemare:
It's not specified in the draft.

stef:
the poc, the rust and my implementation did it like that. i think it comes from the paper. and it makes sense. i'd like to stick with it and rather specify it in the spec, than to drop it.

bytemare:
Why do you pre-hash it before hmac ?

bytemare:
My PR comes from a conversation with @kevinlewi and @chris-wood, and we couldn't explain where this came from.

I think I found it. The V1 says
"
* The explicit concatenation of elements under transcript2 and transcript3 can be replaced with hashed values of these elements, or their combinations, using a collision-resistant hash (e.g., as in the transcript-hash of TLS 1.3 [RFC8446]). 
"

but this section is now absent.

stef:
interesting. i looked in the original paper and it is not there. there is one benefit for prehashing, the fact that the server does not have to preserve all the inputs for the hmac between the create-credential-response and the explicit authentication where the server verifies the km3 mac of the user.

bytemare:
As Transcript3 is defined, it is transcript2 + server mac, and can be assembled at the KE2.
Keeping a hash of Transcript3 reduces the size of the server's state, indeed.

bytemare:
From a security perspective, pre-hashing is a weaker construction, as it relies on the collision resistance of the pre-hashing function. Not hashing protects against a theoretical attack, I don't know if this has been considered in our case.

bytemare:
Actually, the server can compute the client_mac at KE2 as well. It doesn't actually need to preserve anything except the client_mac.

So the server does all its signing in KE2, the client all its signing in KE3, none need to carry any trace of transcript2 and transcripts. Hence, no pre-hashing needed.

stef:
indeed with the elimination of info3/einfo3 this makes sense.

chris-wood:
@bytemare can we close this given that #147 landed?

bytemare:
Yes, of course.

PR #147, opened 2021-02-14
Unify key derivation and confirmation transcripts.

### LINKED ISSUE #108: 3DH transcript is incomplete
The 3DH transcript used in the key schedule [currently includes only the nonces and identity](https://cfrg.github.io/draft-irtf-cfrg-opaque/draft-irtf-cfrg-opaque.html#section-4.2.2.2). This means the contents of info1, info2, and the core OPAQUE messages are not mixed into encryption or MAC keys, which seems less than idea. They are included in the transcript that is covered by the MAC, but I don't see why wouldn't *also* include them in the key schedule. 

@hugokraw, can this be improved?

chris-wood:
This change unifies the transcripts used across key derivation and confirmation. It allows clients and servers to keep a rolling hash of messages sent and received when running the protocol, thereby minimizing state. It also makes testing and implementation easier, as there are no longer two different transcripts which differ in context.

Closes #108 and #142.

cc @bytemare, @stef


bytemare:
That's very interesting, well spotted!

chris-wood:
> But do you hash in hmac because you already hash for info ?

I'm not sure I understand this question 🤔

draft-irtf-cfrg-opaque.md
```diff
@@ -1005,13 +1014,13 @@ IKM = epkU^eskS || epkU^skS || pkU^eskS
 #### OPAQUE-3DH Encryption and Key Confirmation {#3dh-core}
 
 Clients and servers use keys Km2 and Km3 in computing KE2.mac and KE3.mac,
-respectively. These values are computed as HMAC(mac_key, transcript), where
-mac_key and transcript are as follows:
+respectively. These values are computed as HMAC(mac_key, Hash(info || transcript)),
```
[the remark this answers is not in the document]
chris-wood: `suffix` would be a fine name for this, I think?

bytemare:
If I'm not mistaken, `hmac()` doesn't need its input to be hashed beforeheand. Please correct me if I'm wrong.

Here, you build upon the running hash used for `info` and use this state for `hmac()`. `info` needs to hash, so reusing this existing state is beneficial. My question is whether you hash the transcript for `hmac()` so you can use/piggyback on `info`'s running hash.

chris-wood:
> If I'm not mistaken, hmac() doesn't need its input to be hashed beforeheand. Please correct me if I'm wrong.

That's right -- it doesn't. 

> Here, you build upon the running hash used for info and use this state for hmac(). info needs to hash, so reusing this existing state is beneficial. My question is whether you hash the transcript for hmac() so you can use/piggyback on info's running hash.

I *think* I understand you (it would be helpful if you pointed to specific text). I did this because HMAC implementations always hash the input, which is treated as an opaque string. HMAC APIs don't allow one to "pass in" existing hash state, e.g., the running hash context for `info`. 

So, yes, the hash is redundant, but it's there because of API limitations. Does that help clarify?

draft-irtf-cfrg-opaque.md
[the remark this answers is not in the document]
bytemare: sounds good!

draft-irtf-cfrg-opaque.md
```diff
@@ -976,16 +976,25 @@ handshake_encrypt_key =
...
 
-See {{identities}} for more information about identities client_identity and server_identity.
+where `client_preamble` and `server_preamble` are computed as:
+
+~~~
+client_preamble = I2OSP(len(client_identity), 2) || client_identity
+server_preamble = I2OSP(len(server_identity), 2) || server_identity
+~~~
+
+`client_transcript` KE1 and `server_transcript` is KE2 excluding KE2.enc_server_info
```
bytemare, line 993: client_transcript **is**  KE1

bytemare:
Yes, thank you :)

bytemare:
Also, I tested and confirm the new vectors!

draft-irtf-cfrg-opaque.md
chris-wood: 👍 good catch

draft-irtf-cfrg-opaque.md
```diff
@@ -976,16 +976,25 @@ handshake_encrypt_key =
...
 
-The HKDF input parameter `info` is computed as follows:
+The Derive-Secret parameter `info` is computed as:
 
 ~~~
-info = "3DH keys" || I2OSP(len(client_nonce), 2) || client_nonce
-                  || I2OSP(len(server_nonce), 2) || server_nonce
-                  || I2OSP(len(client_identity), 2) || client_identity
-                  || I2OSP(len(server_identity), 2) || server_identity
+info = "3DH" || client_preamble || client_transcript ||
```
kevinlewi, line 982: Can we rename this variable to `preamble` or something, instead of keeping it as `info`? It seems to be called `preamble` in the HMQV/SIGMA-I sections, and `info` is already an overloaded term here, since it could be confused with HKDF info, `client_info`, `enc_server_info `, etc. :)

draft-irtf-cfrg-opaque.md
kevinlewi: The new `client_transcript` variable only appears twice. Can we just rename it as `KE1` in the two occurrences it appears in, rather than introducing a new variable?

Also, same thing with the `server_transcript` variable. I am wary of introducing all of these new terms that aren't accompanied by a struct definition. Perhaps we could abstract away the struct definition of `KE2` to denote an "inner" struct which captures these values that we want to represent as `server_transcript` here? So in other words, something like:

```
struct {
  CredentialResponse response;
  uint8 server_nonce[32];
  uint8 server_keyshare[Npk];
} InnerKE2;

struct {
  InnerKE2 inner_ke2;
  opaque enc_server_info<0..2^16-1>;
  uint8 mac[Nh];
} KE2;
```

And then just use `InnerKe2` here instead of `server_transcript`.

draft-irtf-cfrg-opaque.md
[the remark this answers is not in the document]
kevinlewi: Agreed with not calling this `transcript`. Also, maybe the clarity of this could be improved, by explicitly writing out the computation of KE2.mac and KE2.mac in an equation form (similarly to how the `auth_tag` computation is written for the envelope):

- `Ke2.mac = HMAC(Km2, concat(info, KE2.enc_server_info))`
- `Ke3.mac = HMAC(Km3, concat(info, KE2.enc_server_info, KE2.mac))`

That way, you don't even need to introduce the `transcript` or `suffix` variable :)

kevinlewi (commented):
Providing some initial suggestions after reading through the changes, which look reasonable to me.

I would also like to try to implement this and see if the test vectors match, and also verify whether or not this is actually a simplification of the implementation, but thought it would be good to leave these suggestions here first. But no need to block on me completing the implementation portion.

chris-wood:
Great suggestions, @kevinlewi! I applied them. Please have another look?

kevinlewi:
Looks great! Also verified the test vectors.

kevinlewi (approved):

PR #148, opened 2021-02-15
Simplify static DH text.

### LINKED ISSUE #131: Static DH oracle text
We should ensure that we acknowledge this type of attack in theory, and describe how OPAQUE deals with it. In particular, since each OPRF key is per-user, which means that any reduction in security is per-user, which is unlikely to be a problem in practice.

chris-wood:
Closes #131.


kevinlewi (approved):
LGTM!

Issue #84, opened 2020-11-16
note regarding omitting skU from envelope and deriving it from rwd instead


stef:
in the source there's a commented out part `## Envelope considerations` which mentions that not even `skU` needs to be included in the credentials wrapped in the envelope, instead skU can be derived from `rwd`. there's a note that this part should be "brought back after updating". if this is indeed the case, do not forget to change 
> Applications MUST include skU in secret_credentials
in the {#protocol-messages} section.

chris-wood:
Yeah, we need to remove that commented out text. I'll file a PR for that.

stef:
why do you want to remove it instead of bringing it back? i think this part makes sense.

chris-wood:
> i think this part makes sense.

I need to think about this more, though my sense is that the bandwidth saved is not be worth the additional complexity. The original I-D had this to say about the issue:

```
   Note (storage/communication efficient EnvU): It is possible to
   shorten EnvU by including PubS in ClrEnv and omitting SecEnv and any
   other value from EnvU.  In this case, the user's key PrivU is derived
   from RwdU; e.g., the value PAD in KEYS is used as a seed for a key
   generation procedure that outputs PrivU (the length of PAD will
   depend on the specific key generation procedure).  Such EnvU consists
   of Nonce, PubS and the HMAC value, resulting in less storage at the
   server and less communication from server to client.  The server can
   further minimize storage space by deriving per-user OPRF keys kU from
   a single global secret key using a PRF, and it can use the same pair
   (PrivS,PubS) for all its users.  In this case, the per-user OPAQUE
   storage consists of Nonce, PubU and HMAC(Khmac; PubS).  While it may
   be "tempting" to omit Nonce for space savings, this can lead to
   vulnerabilities (see below).
```

@hugokraw, what do you think?

hugokraw:
I am sure that there are cases where such a savings will be meaningful and I am sure it will not be the typical case; I can't say much more than that. Frankly, the main reason I wanted to "document" this option is that in Bjorn Hasse's AuCPace paper, the amount of storage and/or communication was the only aspect on which AuCPace had some performance advantage. That paper compared against a very wasteful interpretation of what OPAQUE needed and I wanted to document the fact that one can actually be very minimal in envelope's size. I would prefer to make progress with the specification without entering this level of optimization but if there are potential users of OPAQUE for which these savings are important we could at least leave it as a remark.

kevinlewi:
[...]

PR #154, opened 2021-03-02
New Envelope mode proposal including private key derivation from OPRF output

bytemare:
After discussions with @chris-wood and @kevinlewi and the thread in #84, this PR is a first sketch of a proposal to address these discussions.

Proposal of new envelope modes

1. **Drop Base/CustomIdentifiers**
CleartextCredentials always contain pks, idu and ids. If not supplied by the client, idu and ids MUST default to the public keys.
2. Introduce two new modes
    a. **Internal mode**
    The client keys are generated internally from a seed. (also called "seed mode", "encryption-less mode", etc. in different issues and conversations). No encrypted creds in the envelope. Applications are not expected to know how to deal with key generation.
    b. **External mode**
    The client keys are provided by the application, and OPAQUE encrypts the private key into the encrypted creds in the envelope.

Isolating the Envelope description and mode, as well as the rearrangements, are my own suggestions and not part of the discussions.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -306,7 +306,7 @@ those credentials, recovers them using the client's password, and subsequently u
 them as input to an AKE protocol.
 
 Currently, the most widely deployed PKI-free aPAKE is SRP {{?RFC2945}}, which is
-vulnerable to pre-computation attacks, lacks proof of security and is less efficient
+vulnerable to pre-computation attacks, lacks proof of security, and is less efficient
 relative to OPAQUE. Moreover, SRP requires a ring as it mixes addition and
 multiplication operations, and thus does not work over plain elliptic curves. OPAQUE
 is therefore a suitable replacement for applications that use SRP.
@@ -331,8 +331,7 @@ operations, roles, and behaviors of OPAQUE:
   `concat(0x01, 0x0203, 0x040506) = 0x010203040506`.
 - random(n): Generate a cryptographically secure pseudorandom byte string of length `n` bytes.
 - xor(a,b): Apply XOR to byte strings. For example, `xor(0xF0F0, 0x1234) = 0xE2C4`.
-  It is an error to call this function with two arguments of unequal
-  length.
+  It is an error to call this function with two arguments of unequal length.
 - ct_equal(a, b): Return `true` if `a` is equal to `b`, and false otherwise.
   This function is constant-time in the length of `a` and `b`, which are assumed to be of equal length, irrespective of the values `a` or `b`.
 
@@ -356,8 +355,7 @@ OPAQUE relies on the following protocols and primitives:
     output element `Z`.
   - Finalize(x, r, Z): Finalize the OPRF evaluation using input `x`,
     random scalar `r`, and evaluation output `Z`, yielding output `y`.
-  - DeriveKeyPair(seed): Derive a private and public key pair deterministically
-    from a seed.
+  - DeriveKeyPair(seed): Derive a private and public key pair deterministically from a seed.
   - SerializedElement: A serialized OPRF group element, a byte array of fixed
     length.
   - SerializedScalar: A serialized OPRF scalar, a byte array of fixed length.
@@ -393,6 +391,294 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
 We let `Npk` and `Nsk` denote the size of public and private keys, respectively,
 used in the AKE.
 
+- Deterministic AKE Key Generation Functions:
+  - RecoverPublicKey(private_key): Recover the public key related to the input private key.
+  - DeriveAkeKeyPair(seed): Derive a private and public key pair deterministically from the seed.
+
+Random nonces used in this protocol are of length `Nn = 32` bytes.
+
+We define `nil` as a lack of value.
+
+# Client Credential Storage {#client-credential-storage}
+
+OPAQUE makes use of a structure `Envelope` to manage client credentials. This envelope holds information about its format and content for the client to obtain its authentication material.
+
+OPAQUE allows applications to either provide custom client private and public keys for authentication or to generate them internally, making the application oblivious to the client's private key. Each public and private key value is an opaque byte string, specific to the AKE protocol in which OPAQUE is instantiated.
+
+These two options are defined as the `Internal Key` and `External Key` modes.  See {{envelope-modes}} for their specifications.
+
+Credential material is used to pin key material to identities if desired. If no identity is given for a party, its value MUST default to its public key. The following types of credentials are hereby considered:
+
+- client_private_key: The encoded client private key for the AKE protocol.
+- client_public_key: The encoded client public key for the AKE protocol.
+- server_public_key: The encoded server public key for the AKE protocol.
+- client_identity: The client identity. This is an application-specific value, e.g., an e-mail address or normal account name. If not specified, it defaults to the client's public key.
+- server_identity: The server identity. This is typically a domain name, e.g., example.com.  If not specified, it defaults to the server's public key.
+  See {{identities}} for information about this identity.
+
+These values are used in the `CleartextCredentials` structure as follows:
+
+~~~
+struct {
+ opaque server_public_key[Npk];
+ opaque server_identity<1..2^16-1>;
+ opaque client_identity<1..2^16-1>;
+} CleartextCredentials;
+~~~
+
+The function CreateCleartextCredentials implements this logic:
+
+~~~
+CreateCleartextCredentials(server_public_key, client_public_key, server_identity, client_identity)
+
+Input:
+- server_public_key, The encoded server public key for the AKE protocol
+- client_public_key, The encoded client public key for the AKE protocol
+- server_identity, The optional encoded server identity, which is nil if not specified
+- client_identity, The optional encoded client identity, which is nil if not specified
+
+Output:
+- cleartext_credentials, a CleartextCredentials structure
+
+Steps:
+1. if server_identity == nil
+2.  server_identity = server_public_key
+3. if client_identity == nil
+4.  client_identity = client_public_key
+5. Create CleartextCredentials cleartext_credentials with (server_public_key, server_identity, client_identity)
+6. Output cleartext_credentials
+~~~
+
+During protocol execution, the identity values can be stored in an implementation-specific `Credentials` object with names matching the values.
+
+~~~
+struct {
+ opaque server_identity;
+ opaque client_identity;
+} Credentials;
+~~~
+
+## Envelope Structure {#envelope-structure}
+
+A client `Envelope` is constructed independently from the `EnvelopeMode`, but offers an `InnerEnvelope` entry that modes can use as they see fit.
+`Envelope` is constructed as follows:
+
+~~~
+struct {
+ EnvelopeMode mode;
+ opaque nonce[Nn];
+ opaque auth_tag[Nm];
+ InnerEnvelope inner_env;
+} Envelope;
+~~~
+
+mode : The `EnvelopeMode` used for the `Envelope`.
+
+nonce : A unique nonce of length `Nn` used to protect this Envelope.
+
+auth_tag : Authentication tag protecting the contents of the envelope, covering `EnvelopeMode`, envelope nonce, `InnerEnvelope` and `CleartextCredentials`.
+
+inner_env : A mode dependent `InnerEnvelope` structure. See {{envelope-modes}} for its specifications.
+
+The size of the serialized envelope is denoted `Ne` and varies based on the mode. The exact value for `Ne` is specified in {{internal-mode}} and {{external-mode}}.
+
+The `CreateEnvelope` function allows a client to create its `Envelope` at registration.
+
+~~~
+CreateEnvelope(prk, server_public_key, client_private_key, creds)
+
+Parameter:
+- mode, the EnvelopeMode mode
+
+Input:
+- prk, pseudo-random key
+- server_public_key, The encoded server public key for the AKE protocol
+- client_private_key, The encoded client private key for the AKE protocol
+- server_identity, The encoded server identity
+- client_identity, The encoded client identity
+
+Output:
+- envelope, the client's `Envelope` structure
+- client_public_key, the client's AKE public key
+- masking_key, a key used by the server to preserve the confidentiality of the envelope during login
+- export_key, an additional key
+
+Steps:
+1. envelope_nonce = random(Nn)
+2. auth_key = Expand(prk, concat(envelope_nonce, "AuthKey"), Nh)
+3. export_key = Expand(prk, concat(envelope_nonce, "ExportKey"), Nh)
+4. masking_key = Expand(prk, "MaskingKey", Nh)
+5. inner_env, client_public_key = BuildInnerEnvelope(prk, envelope_nonce, client_private_key)
+6. cleartext_creds = CreateCleartextCredentials(server_public_key, client_public_key, server_identity, client_identity)
+7. auth_tag = MAC(auth_key, concat(mode, envelope_nonce, inner_env, cleartext_creds))
+8. Create Envelope envelope with (mode, envelope_nonce, auth_tag, inner_env)
+9. Output (envelope, client_public_key, masking_key, export_key)
+~~~
+
+The `RecoverEnvelope` function allows a client to recover its `Envelope` at authentication.
+
+~~~  
+RecoverEnvelope(prk, server_public_key, creds, envelope)
+
+Parameter:
+- mode, the EnvelopeMode mode
+
+Input:
+- prk, pseudo-random key
+- server_public_key, The encoded server public key for the AKE protocol
+- creds, a Credentials structure
+- envelope, the client's `Envelope` structure
+  
+Output:
+- client_private_key, The encoded client private key for the AKE protocol
+- export_key, an additional key
+
+Steps:
+1. auth_key = Expand(prk, concat(envelope.nonce, "AuthKey"), Nh)
+2. export_key = Expand(prk, concat(envelope.nonce, "ExportKey", Nh)
+3. client_private_key, client_public_key = RecoverKeys(prk, envelope.nonce, envelope.inner_env)
+4. cleartext_creds = CreateCleartextCredentials(server_public_key, client_public_key, creds.server_identity, creds.client_identity)
+5. expected_tag = MAC(auth_key, concat(mode, envelope_nonce, inner_env, cleartext_creds))
+6. If !ct_equal(envelope.auth_tag, expected_tag),
+     raise MacError
+7. Output (client_private_key, export_key)
+~~~
+
+## Envelope Modes {#envelope-modes}
+
+The `EnvelopeMode` specifies the structure and encoding of the corresponding `InnerEnvelope`. This document specifies the values of the two aformentionned modes:
+
+~~~
+enum {
+ internal(1),
+ external(2),
+ (255)
+} EnvelopeMode;
+~~~
+
+Each `EnvelopeMode` defines its own `InnerEnvelope` structure and must implement the following interface:
+- `inner_env, client_public_key = BuildInnerEnvelope(prk, nonce, client_private_key)`: build and return the mode's `InnerEnvelope` structure and the client's public key.
+- `client_private_key, client_public_key = RecoverKeys(prk, nonce, inner_env)`: recover and return the client's private and public keys for the AKE protocol.
+
+### Internal Key mode {#internal-mode}
+
+In this mode, the client's private and public keys are deterministically derived from the OPRF output. 
+
+With the internal key mode the `EnvelopeMode` value MUST be `internal` and the `InnerEnvelope` is empty, and the size `Ne` of the serialized `Envelope` is 1 + Nn + Nm.
+
+To generate the private key OPAQUE-3DH implements `DeriveAkeKeyPair(seed)` as follows:
+
+~~~
+DeriveAkeKeyPair(seed)
+
+Parameter:
+- dst, domain separation tag for HashToScalar set to "OPAQUE-HashToScalar"
+
+Input:
+- seed, pseudo-random byte sequence used as a seed
+
+Output:
+- private_key, a private key
+- public_key, the associated public key
+
+Steps:
+1. private_key = HashToScalar(seed, dst)
+2. public_key = private_key * G
+3. Output (private_key, public_key)
+~~~
+
+HashToScalar(msg, dst): Hash arbitrary input to a scalar in the field of the AKE group as defined in {{I-D.irtf-cfrg-voprf}}, except that dst = "OPAQUE-HashToScalar".
+
+~~~
+BuildInnerEnvelope(prk, nonce, client_private_key)
+
+Input:
+- prk, pseudo-random key
+- nonce, a unique nonce of length `Nn`
+- client_private_key, empty value. Not used in this function, it only serves to comply with the API
+
+Output:
+- inner_env, nil value (serves to comply with the API)
+- client_public_key, the client's AKE public key
+
+Steps:
+1. seed = Expand(prk, concat(nonce, "PrivateKey"), Nh)
+2. _, client_public_key = DeriveAkeKeyPair(seed)
+3. Output (nil, client_public_key)
+~~~
+
+~~~
+RecoverKeys(prk, nonce, inner_env)
+
+Input:
+- prk, pseudo-random key
+- nonce, a unique nonce of length `Nn`
+- inner_env, an InnerEnvelope structure. Not used in this function, it only serves to comply with the API
+
+Output:
+- client_private_key, The encoded client private key for the AKE protocol
+- client_public_key, The encoded client public key for the AKE protocol
+
+Steps:
+1. seed = Expand(prk, concat(nonce, "PrivateKey"), Nh)
+2. client_private_key, client_public_key = DeriveAkeKeyPair(seed)
+4. Output (client_private_key, client_public_key)
+~~~
+
+### External Key mode {#external-mode}
+
+With the external key mode the `EnvelopeMode` value MUST be `external`, and the size `Ne` of the serialized `Envelope` is 1 + Nn + Nm + Nsk.
+
+In this mode, the client's public key is recovered using the private key. An encryption key is generated from the hardened OPRF output and used to encrypt the client's private key, which is then stored encrypted in the `InnerEnvelope`. This allows for applications that prefer to use custom keys.
+
+~~~
+struct {
+ opaque encrypted_creds[Nsk];
+} InnerEnvelope;
+~~~
+
+encrypted_creds : Encrypted client_private_key. Authentication of this field is ensured with the `AuthTag` in the envelope that covers this `InnerEnvelope`.
+
+~~~
+BuildInnerEnvelope(prk, nonce, client_private_key)
+
+Input:
+- prk, pseudo-random key
+- nonce, a unique nonce of length `Nn`
+- client_private_key, The encoded client private key for the AKE protocol
+
+Output:
+- inner_env, an InnerEnvelope structure
+- client_public_key, The encoded client public key for the AKE protocol
+
+Steps:
+1. pseudorandom_pad = Expand(prk, concat(nonce, "Pad"), len(client_private_key))
+2. encrypted_creds = xor(client_private_key, pseudorandom_pad)
+3. Create InnerEnvelope inner_env with encrypted_creds
+4. client_public_key = RecoverPublicKey(client_private_key)
+5. Output (inner_env, client_public_key)
+~~~
+
+~~~
+RecoverKeys(prk, nonce, inner_env)
+
+Input:
+- prk, pseudo-random key
+- nonce, a unique nonce of length `Nn`
+- inner_env, an InnerEnvelope structure
+
+Output:
+- client_private_key, The encoded client private key for the AKE protocol
+- client_public_key, the client's AKE public key
+
+Steps:
+1. encrypted_creds = inner_env.encrypted_creds
+2. pseudorandom_pad = Expand(prk, concat(nonce, "Pad"), len(encrypted_creds))
+3. client_private_key = xor(encrypted_creds, pseudorandom_pad)
+4. client_public_key = RecoverPublicKey(client_private_key)
+5. Output (client_private_key, client_public_key)
+~~~
+
 # Offline Registration {#offline-phase}
 
 Registration is executed between a client C and a
@@ -407,8 +693,8 @@ Clients MUST NOT use the same key pair (client_private_key, client_public_key) f
 ## Setup Phase {#setup-phase}
 
 In a setup phase, C chooses its password, and S chooses its own pair of private-public
-AKE keys (server_private_key, server_public_key) for use with the AKE, along with a Nh-byte oprf_seed. S can use
-the same pair of keys with multiple clients, and can opt to use multiple seeds (so long as they are
+AKE keys (server_private_key, server_public_key) for use with the AKE, along with an Nh-byte oprf_seed. S can use
+the same pair of keys with multiple clients and can opt to use multiple seeds (so long as they are
 kept consistent for each client). These steps can happen offline, i.e., before the registration phase.
 
 Once complete, the registration process proceeds as follows.
@@ -428,7 +714,7 @@ Once complete, the registration process proceeds as follows.
                                response
                       <-------------------------
 
- (record, export_key) = FinalizeRequest(password, creds, blind, response)
+ (record, export_key) = FinalizeRequest(client_private_key, password, blind, response)
 
                                 record
                       ------------------------->
@@ -440,105 +726,6 @@ Both client and server MUST validate the other party's public key before use.
 See {{validation}} for more details.
 
 Upon completion, S stores C's credentials for later use.
-
-## Credential Storage {#credential-storage}
-
-OPAQUE makes use of a structure `Envelope` to store client credentials.
-The `Envelope` structure embeds the following types of credentials:
-
-- client_private_key: The encoded client private key for the AKE protocol.
-- server_public_key: The encoded server public key for the AKE protocol.
-- client_identity: The client identity. This is an application-specific value, e.g., an e-mail
-  address or normal account name.
-- server_identity: The server identity. This is typically a domain name, e.g., example.com.
-  See {{identities}} for information about this identity.
-
-Each public and private key value is an opaque byte string, specific to the AKE
-protocol in which OPAQUE is instantiated. For example, if used as raw public keys
-for TLS 1.3 {{?RFC8446}}, they may be RSA or ECDSA keys as per {{?RFC7250}}.
-
-These credentials are incorporated in the `SecretCredentials` and `CleartextCredentials` structs,
-depending on the mode set by the value of `EnvelopeMode`:
-
-~~~
-enum {
-  base(1),
-  custom_identifier(2),
-  (255)
-} EnvelopeMode;
-~~~
-
-The `base` mode defines `SecretCredentials` and `CleartextCredentials` as follows:
-
-~~~
-struct {
-  opaque client_private_key[Nsk];
-} SecretCredentials;
-
-struct {
-  opaque server_public_key[Npk];
-} CleartextCredentials;
-~~~
-
-The `custom_identifier` mode defines `SecretCredentials` and `CleartextCredentials` as follows:
-
-~~~
-struct {
-  opaque client_private_key[Nsk];
-} SecretCredentials;
-
-struct {
-  opaque server_public_key[Npk];
-  opaque client_identity<0..2^16-1>;
-  opaque server_identity<0..2^16-1>;
-} CleartextCredentials;
-~~~
-
-These credentials are embedded into the following `Envelope` structure with
-encryption and authentication.
-
-~~~
-struct {
-  EnvelopeMode mode;
-  opaque nonce[32];
-  opaque encrypted_creds[Nsk];
-} InnerEnvelope;
-
-struct {
-  InnerEnvelope inner_env;
-  opaque auth_tag[Nm];
-} Envelope;
-~~~
-
-mode
-: The `EnvelopeMode` value. This MUST be one of `base` or `custom_identifier`.
-
-nonce
-: A unique 32-byte nonce used to protect this Envelope.
-
-encrypted_creds
-: Encoding of encrypted and authenticated `SecretCredentials`.
-
-auth_tag
-: Authentication tag protecting the contents of the envelope,
-covering `InnerEnvelope` and `CleartextCredentials`.
-
-The full procedure for constructing `Envelope` and `InnerEnvelope` from
-`SecretCredentials` and `CleartextCredentials` is described in {{finalize-request}}.
-Note that only `SecretCredentials` are stored in the `Envelope` (in encrypted form).
-
-The `EnvelopeMode` value is specified as part of the configuration (see {{configurations}}).
-
-Credential information corresponding to the configuration-specific mode,
-along with the client public key `client_public_key` and private key `client_private_key`,
-are recommended to be stored in a `Credentials` object with the following named fields:
-
-- `client_private_key`, the client's private key
-- `client_public_key`, the client's public key corresponding to `client_private_key`
-- `client_identity`, an optional client identity (present only in the `custom_identifier` mode)
-- `server_identity`, an optional server identity (present only in the `custom_identifier` mode)
-
-Note that the total size of the Envelope is equal to Nsk + Nh + 33 bytes.
 
 ## Registration Messages
 
@@ -627,17 +814,13 @@ Steps:
 ### FinalizeRequest {#finalize-request}
 
 ~~~
-FinalizeRequest(password, creds, blind, response)
-
-Parameters:
-- params, the MHF parameters established out of band
-- mode, the InnerEnvelope mode
-- Nx, the output size of the Extract function
+FinalizeRequest(client_private_key, password, blind, response)
 
 Input:
+- client_private_key, the client's private key. In the internal key mode, this is empty.
 - password, an opaque byte string containing the client's password
 - creds, a Credentials structure
-- blind, an OPRF scalar value
+- blind, the OPRF scalar value used for blinding
 - response, a RegistrationResponse structure
 
 Output:
@@ -646,33 +829,21 @@ Output:
 
 Steps:
 1. y = Finalize(password, blind, response.data)
-2. envelope_nonce = random(32)
-3. prk = Extract("", Harden(y, params))
-4. Create SecretCredentials secret_creds with creds.client_private_key
-5. Create CleartextCredentials cleartext_creds with response.server_public_key
-   and custom identifiers creds.client_identity and creds.server_identity if
-   mode is custom_identifier
-6. pseudorandom_pad = Expand(prk, concat(envelope_nonce, "Pad"), len(secret_creds))
-7. auth_key = Expand(prk, concat(envelope_nonce, "AuthKey"), Nh)
-8. export_key = Expand(prk, concat(envelope_nonce, "ExportKey"), Nh)
-9. masking_key = Expand(prk, "MaskingKey", Nh)
-10. encrypted_creds = xor(secret_creds, pseudorandom_pad)
-11. Create InnerEnvelope inner_env
-      with (mode, envelope_nonce, encrypted_creds)
-12. auth_tag = MAC(auth_key, concat(inner_env, cleartext_creds))
-13. Create Envelope envelope with (inner_env, auth_tag)
-14. Create RegistrationUpload record with (creds.client_public_key, masking_key, envelope)
-15. Output (record, export_key)
+2. prk = Extract("", Harden(y, params))
+3. envelope, client_public_key, masking_key, export_key = CreateEnvelope(prk, response.server_public_key, client_private_key, creds)
+4. Create RegistrationUpload record with (client_public_key, masking_key, envelope)
+5. Output (record, export_key)
 ~~~
-
-The inputs to Extract and Expand are as specified in {{dependencies}}.
 
 See {{online-phase}} for details about the output export_key usage.
 
 Upon completion of this function, the client MUST send `record` to the server.
 
-The server then directly stores the `record` object as the credential file for each client. Note that
-the values `oprf_seed` and `server_private_key` from the server's setup phase must also be persisted.
+### Finalize Registration {#finalize-registration}
+
+The server stores the `record` object as the credential file for each client along with the associated `credential_identifier`
+and `client_identity` (if different).
+Note that the values `oprf_seed` and `server_private_key` from the server's setup phase must also be persisted.
 
 # Online Authenticated Key Exchange {#online-phase}
 
@@ -688,7 +859,7 @@ shared secret key.
 
 OPAQUE produces two outputs: a session secret and an export key. The export key may be used
 for additional application-specific purposes, as outlined in {{export-key-usage}}.
-The output `export_key` MUST NOT be used in any way before the HMAC value in the
+The output `export_key` MUST NOT be used in any way before the MAC value in the
 envelope is validated. See {{envelope-encryption}} for more details about this requirement.
 
 ## Credential Retrieval
@@ -731,8 +902,8 @@ data
 ~~~
 struct {
     SerializedElement data;
-    opaque masking_nonce[32];
-    opaque masked_response[Npk + Nsk + Nh + 33];
+    opaque masking_nonce[Nn];
+    opaque masked_response[Npk + Ne];
 } CredentialResponse;
 ~~~
 
@@ -795,7 +966,7 @@ Steps:
 2. Z = Evaluate(oprf_key, request.data)
 3. masking_nonce = random(32)
 4. credential_response_pad = Expand(record.masking_key,
-     concat(masking_nonce, "CredentialResponsePad"), Npk + Nsk + Nh + 33)
+     concat(masking_nonce, "CredentialResponsePad"), Npk + Ne)
 5. masked_response = xor(credential_response_pad, concat(server_public_key, record.envelope))
 6. Create CredentialResponse response with (Z, masking_nonce, masked_response)
 7. Output response
@@ -804,7 +975,7 @@ Steps:
 In the case of a record that does not exist, the server invokes the CreateCredentialResponse
 function where the record argument is configured so that:
 - record.masking_key is set to a random byte string of length Nh, and
-- record.envelope is set to the byte string consisting only of zeros, of length Nsk + Nh + 33
+- record.envelope is set to the byte string consisting only of zeros, of length Ne
 
 Note that the responses output by either scenario are indistinguishable to an adversary
 that is unable to guess the registered password for the client corresponding to credential_identifier.
@@ -812,16 +983,13 @@ that is unable to guess the registered password for the client corresponding to 
 #### RecoverCredentials {#recover-credentials}
 
 ~~~
-RecoverCredentials(password, blind, response)
-
-Parameters:
-- params, the MHF parameters established out of band
-- Nx, the output size of the Extract function
+RecoverCredentials(password, blind, response, creds)
 
 Input:
 - password, an opaque byte string containing the client's password
 - blind, an OPRF scalar value
 - response, a CredentialResponse structure
+- creds, a Credentials structure
 
 Output:
 - client_private_key, the client's private key for the AKE protocol
@@ -833,22 +1001,10 @@ Steps:
 2. prk = Extract("", Harden(y, params))
 3. masking_key = Expand(prk, "MaskingKey", Nh)
 4. credential_response_pad = Expand(masking_key,
-     concat(response.masking_nonce, "CredentialResponsePad"), Npk + Nsk + Nh + 33)
+     concat(response.masking_nonce, "CredentialResponsePad"), Npk + Ne)
 5. concat(server_public_key, envelope) = xor(credential_response_pad, response.masked_response)
-6. contents = envelope.contents
-7. envelope_nonce = contents.nonce
-8. pseudorandom_pad =
-    Expand(prk, concat(envelope_nonce, "Pad"), len(contents.encrypted_creds))
-9. auth_key = Expand(prk, concat(envelope_nonce, "AuthKey"), Nh)
-10. export_key = Expand(prk, concat(envelope_nonce, "ExportKey"), Nh)
-11. Create CleartextCredentials cleartext_creds with server_public_key
-   and custom identifiers creds.client_identity and creds.server_identity if mode is
-   custom_identifier
-12. expected_tag = MAC(auth_key, concat(contents, cleartext_creds))
-13. If !ct_equal(envelope.auth_tag, expected_tag),
-    raise DecryptionError
-14. secret_creds = xor(contents.encrypted_creds, pseudorandom_pad)
-15. Output (secret_creds.client_private_key, server_public_key, export_key)
+6. client_private_key, export_key = RecoverEnvelope(prk, server_public_key, creds, envelope)
+7. Output (client_private_key, response.server_public_key, export_key)
 ~~~
 
 ## AKE Instantiations {#instantiations}
@@ -922,7 +1078,7 @@ The three messages for OPAQUE-3DH are described below.
 ~~~
 struct {
   CredentialRequest request;
-  uint8 client_nonce[32];
+  uint8 client_nonce[Nn];
   opaque client_info<0..2^16-1>;
   uint8 client_keyshare[Npk];
 } KE1;
@@ -932,7 +1088,7 @@ request
 : A `CredentialRequest` generated according to {{create-credential-request}}.
 
 client_nonce
-: A fresh 32-byte randomly generated nonce.
+: A fresh randomly generated nonce of length `Nn`.
 
 client_info
 : Optional application-specific information to exchange during the protocol.
@@ -945,7 +1101,7 @@ prime order group.
 struct {
   struct {
     CredentialResponse response;
-    uint8 server_nonce[32];
+    uint8 server_nonce[Nn];
     uint8 server_keyshare[Npk];
   } inner_ke2;
   opaque enc_server_info<0..2^16-1>;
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -385,145 +385,329 @@ variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}.
...
+- `server_identity`, an optional server identity (present only in the `custom_identifier` mode)
 
- (record, export_key) = FinalizeRequest(password, creds, blind, response)
+<!-- TODO(Daniel):
+I don't think the second part of the following is still relevant given the 3DH setup.
+-->
 
-                                record
-                      ------------------------->
+Each public and private key value is an opaque byte string, specific to the AKE protocol in which OPAQUE is instantiated. For example, if used as raw public keys for TLS 1.3 {{?RFC8446}}, they may be RSA or ECDSA keys as per {{?RFC7250}}.
```
bytemare, line 428: > For example, if used as raw public keys for TLS 1.3 {{?RFC8446}}, they may be RSA or ECDSA keys as per {{?RFC7250}}.
This is not relevant anymore in that part of the document.

draft-irtf-cfrg-opaque.md
```diff
@@ -385,145 +385,329 @@ variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}.
...
+is that which is associated with the OPAQUE configuration (see {{configurations}}).
+
+<!-- TODO(Daniel):
+todo: explain rationale
+
+- this document defines opaque with 3DH
+- 3DH is over POG, i.e. can leverage internal hash-to-scalar Key gen
+- no need for the application to deal with key generation, no encrypted creds
+- for compatibility with other AKEs, external mode that allows arbitrary key kind importation with encrypted credentials.
+-->
```
bytemare, line 533: I will do a commit on this later

draft-irtf-cfrg-opaque.md
```diff
@@ -385,145 +385,329 @@ variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}.
...
+mode
+: The `EnvelopeMode` value. This MUST be `internal`.
+
+nonce
+: A unique 32-byte nonce used to protect this Envelope.
+
+<!-- TODO(Daniel):
+1. Describe KeyGenFromSeed(seed) in general
+2. Describe key gen = HashToScalar(prk, "OPAQUE04-HashToScalar") in the 3DH setup
+3. Define HashToScalar and ScalarBaseMult
```
bytemare, line 558: @chris-wood I think we should define the HashToScalar similarly to what has been done in VOPRF. What do you think?

draft-irtf-cfrg-opaque.md
```diff
@@ -385,145 +385,329 @@ variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}.
...
+ } InnerEnvelope;
+~~~
+mode
+: The `EnvelopeMode` value. This MUST be `internal`.
+
+nonce
+: A unique 32-byte nonce used to protect this Envelope.
+
+<!-- TODO(Daniel):
+1. Describe KeyGenFromSeed(seed) in general
```
bytemare, line 556: I'm open to suggestions on how to best formulate this.
I'm thinking of something around this

> KeyGenFromSeed(seed): expand the seed to full private/public generation in the AKE setup

kevinlewi: Can't this just be equal to hash_to_field with a specific DST?

draft-irtf-cfrg-opaque.md
kevinlewi: Either hash_to_field or HashToScalar from VOPRF (may be cleaner to just specify hash_to_field from the h2c spec)

draft-irtf-cfrg-opaque.md
```diff
@@ -385,145 +385,329 @@ variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}.
...
 
-            (response, oprf_key) = CreateRegistrationResponse(request, server_public_key)
+During protocol execution, they are recommended to be stored in a `Credentials` object with the following named fields:
 
-                               response
-                      <-------------------------
+- `client_private_key`, the client's private key
+- `client_public_key`, the client's public key corresponding to `client_private_key`
+- `client_identity`, an optional client identity (present only in the `custom_identifier` mode)
+- `server_identity`, an optional server identity (present only in the `custom_identifier` mode)
```
kevinlewi, line 422: reference to `custom_identifier` mode which needs to be updated

draft-irtf-cfrg-opaque.md
```diff
@@ -385,145 +385,329 @@ variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}.
...
-  address or normal account name.
-- server_identity: The server identity. This is typically a domain name, e.g., example.com.
-  See {{identities}} for information about this identity.
+Steps:
+1. envelope_nonce = random(32)
+2. prk = HKDF-Extract(envelope_nonce, Harden(y, params))
+3. auth_key = HKDF-Expand(prk, "AuthKey", Nh)
+4. export_key = HKDF-Expand(prk, "ExportKey", Nh)
+5. inner_env, client_public_key = BuildInnerEnvelope(prk, creds)
+6. Create CleartextCredentials cleartext_creds with response.server_public_key and custom identifiers creds.client_identity and creds.server_identity if mode is custom_identifier
```
kevinlewi, line 471: reference to `custom_identifier` mode which needs to be updated

draft-irtf-cfrg-opaque.md
```diff
@@ -385,145 +385,329 @@ variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}.
...
+  
+Output:
+- secret_creds, SecretCredentials
+- export_key, an additional key
+
+Steps:
+1. prk = HKDF-Extract(envelope_nonce, Harden(y, params))
+2. auth_key = HKDF-Expand(prk, "AuthKey", Nh)
+3. export_key = HKDF-Expand(prk, "ExportKey", Nh)
+4. Create CleartextCredentials cleartext_creds with response.server_public_key and custom identifiers creds.client_identity and creds.server_identity if mode is custom_identifier
```
kevinlewi, line 499: reference to `custom_identifier` mode which needs to be updated

draft-irtf-cfrg-opaque.md
```diff
@@ -385,145 +385,329 @@ variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}.
...
+OPAQUE allows applications to either provide custom client private and public keys for authentication, or to generate them internally, making the application oblivious to the client's private key.
 
+These two options are defined as the `Internal Key` and `External Key` modes.  See {{envelope-modes}} for their specifications.
+
+Credential material is used to pin key material to identities if desired. If no identity is given for a party, its value MUST default to its public key. The following types of credentials are hereby considered:
+
+- client_private_key: The encoded client private key for the AKE protocol.
+- server_public_key: The encoded server public key for the AKE protocol.
+- client_identity: The client identity. This is an application-specific value, e.g., an e-mail address or normal account name. If not specified, it defaults to the client's public key.
+- server_identity: The server identity. This is typically a domain name, e.g., example.com.  If not specified, it defaults to the server's public key.
```
kevinlewi, line 401: I like the way this is written! Let's just remove all of the other references to this identity being "optional" in the later texts, and we can clarify that it defaults to the client_public_key and server_public_key.

kevinlewi (changes requested):

draft-irtf-cfrg-opaque.md
bytemare: I think we should describe a generic `KeyGenFromSeed(seed)` here, and then specify that in OPAQUE-3DH we use `hash_to_field` with dst etc.. This way, it stays relevant for other AKEs, that  just need to define their own `KeyGenFromSeed(seed)` implementation.

draft-irtf-cfrg-opaque.md
bytemare: That's a leftover, well spotted!

draft-irtf-cfrg-opaque.md
bytemare: Maybe "redefine but with the same definition as in VOPRF", just to make clear not to use VOPRF's DST for this hash_to_field

chris-wood: I think the implementation should be something like:

```
sk = HashToScalar(seed)
pk = sk * G
```

bytemare: ok, with something like "OPAQUE-HashToScalar" for dst?

draft-irtf-cfrg-opaque.md
chris-wood: (Discussed offline) let's just mimic what the VOPRF draft does for HashToScalar here :)

bytemare:
Line 302-306 would need to be adapted as well:

> OPAQUE consists of two stages: registration and authenticated key exchange.
> In the first stage, a client registers its password with the server and stores
> its encrypted credentials on the server. In the second stage, a client obtains
> those credentials, recovers them using the client's password, and subsequently uses
> them as input to an AKE protocol.

draft-irtf-cfrg-opaque.md
```diff
@@ -1153,12 +1291,12 @@ server_public_key instead) is acceptable.
 The analysis of OPAQUE from {{OPAQUE}} requires the authenticated encryption scheme
 used to produce envelope to have a special property called random key-robustness
 (or key-committing). This specification enforces this property by utilizing
-encrypt-then-HMAC in the construction of the envelope. There is no option to use another
+encrypt-then-MAC in the construction of the envelope. There is no option to use another
```
bytemare, line 1294: @kevinlewi should the modification be applied here as well?

bytemare:
Latest changes include:
- Moving envelope mode and nonce into the envelope and out of the inner envelope. The inner envelope is now purely independent from one mode to the other, and the internal mode uses a "nil" envelope
- In case no custom identifiers are used, the client didn't have access to its public key. It is now sent by the server in the CredentialResponse and integrated with CleartextCredentials and authenticated under the mac
- abstracted nonce length to not have magic numbers laying around in the document


Todo: adapt sending of Pkc only if custom identifiers are not used

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+ opaque client_private_key[Nsk];
+} SecretCredentials;
 
-                               request
-                      ------------------------->
+struct {
+ opaque server_public_key[Npk];
+ opaque client_public_key[Npk];
+ opaque server_identity<0..2^16-1>;
+ opaque client_identity<0..2^16-1>;
```
chris-wood, line 434:
```suggestion
 opaque server_identity<1..2^16-1>;
 opaque client_identity<1..2^16-1>;
```

(Since neither field can be empty)

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+} CleartextCredentials;
+~~~
 
-            (response, oprf_key) = CreateRegistrationResponse(request, server_public_key)
+The function CreateCleartextCredentials implements this logic:
+~~~
```
chris-wood, line 439:
```suggestion

~~~
```

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
 
-Both client and server MUST validate the other party's public key before use.
-See {{validation}} for more details.
+## Envelope Structure {#envelope-structure}
 
-Upon completion, S stores C's credentials for later use. See {{credential-file}}
-for a recommended storage format.
+A client `Envelope` is constructed independently from the `EnvelopeMode`, but offers an `InnerEnvelope` entry that modes can use as they see fit.
+`Envelope` is constructed as follows:
+~~~
```
chris-wood, line 468:
```suggestion

~~~
```

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+struct {
+ EnvelopeMode mode;
+ opaque nonce[Nn];
+ opaque auth_tag[Nh];
```
chris-wood, line 472: Why is `auth_tag` _before_ `inner_env`? It seems like `auth_tag` should cover the entire envelope, right?

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
-The `Envelope` structure embeds the following types of credentials:
+nonce : A unique nonce of length `Nn` used to protect this Envelope.
 
-- client_private_key: The encoded client private key for the AKE protocol.
-- server_public_key: The encoded server public key for the AKE protocol.
-- client_identity: The client identity. This is an application-specific value, e.g., an e-mail
-  address or normal account name.
-- server_identity: The server identity. This is typically a domain name, e.g., example.com.
-  See {{identities}} for information about this identity.
+auth_tag : Authentication tag protecting the contents of the envelope, covering `InnerEnvelope` and `CleartextCredentials`.
```
chris-wood, line 481: This also includes `mode` and `nonce`, right?

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+
+inner_env : A mode dependent `InnerEnvelope` specifying at least the `EnvelopeMode`.
+
+The `CreateEnvelope` function allows a client to create its `Envelope` at registration.
+~~~
+CreateEnvelope(y, pks, creds)
```
chris-wood, line 487:
```suggestion
CreateEnvelope(random_pwd, server_public_key, creds)
```

chris-wood: (here and below!)

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+- pks, the server AKE public key
+
+Output:
+- envelope, the client's `Envelope` structure
+- client_public_key, the client's AKE public key
+- export_key, an additional key
+
+Steps:
+1. envelope_nonce = random(Nn)
+2. prk = Extract(envelope_nonce, Harden(y, params))
```
chris-wood, line 507: @kevinlewi's change removed the nonce from this step, so we can probably lift derivation of `prk` to the function that calls `CreateEnvelope`.

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+3. auth_key = Expand(prk, "AuthKey", Nx)
+4. export_key = Expand(prk, "ExportKey", Nx)
+5. inner_env, client_public_key = BuildInnerEnvelope(prk, creds)
+6. cleartext_creds = CreateCleartextCredentials(response.server_public_key, client_public_key, creds.server_identity, creds.client_identity)
+7. auth_tag = MAC(auth_key, concat(mode, envelope_nonce, inner_env, cleartext_creds))
+8. Create Envelope envelope with (mode, envelope_nonce, auth_tag, inner_env)
+9. Output (envelope, client_public_key, export_key)
+~~~
+
+The `RecoverEnvelope` function allows a client to recover its `Envelope` at authentication. Note that the `creds` structure MUST at least hold the server_public_key and client_public_key values.
```
chris-wood, line 517: The server public key is passed in as a parameter (as part of the CredentialResponse), so why is it needed to be part of `creds`?

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+Each `EnvelopeMode` defines its own `InnerEnvelope` structure and must implement the following interface:
+- `inner_env, client_public_key = BuildInnerEnvelope(prk, creds)`: build and return the `InnerEnvelope` structure and the client's public key.
+- `secret_creds = RecoverSecret(prk, inner_env)`: recover and return the `SecretCredentials` structure.
+
+The inputs to Extract and Expand are as specified in {{RFC5869}}. The underlying hash function
+is that which is associated with the OPAQUE configuration (see {{configurations}}).
+
+### Internal Key mode {#internal-mode}
+
+In this mode, the client's private and public keys are deterministically derived from the OPRF output. For OPAQUE-3DH, this is done by hashing the OPRF output to a scalar in the corresponding prime order group to produce a private key, and then multiplying that scalar by the fixed generator `G` to produce the public key.
```
chris-wood, line 567:
```suggestion
In this mode, the client's private and public keys are deterministically derived from the OPRF output. 
```

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
 
+With the internal key mode the `EnvelopeMode` value MUST be `internal` and the `InnerEnvelope` is empty of nil. Indeed, this mode does not need it.
```
chris-wood, line 569:
```suggestion
With the internal key mode the `EnvelopeMode` value MUST be `internal` and the `InnerEnvelope` is empty.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
-struct {
-  opaque client_private_key[Nsk];
-} SecretCredentials;
+BuildInnerEnvelope(prk, creds)
 
-struct {
-  opaque server_public_key[Npk];
-} CleartextCredentials;
+Parameters:
+- group, the AKE group to derive the keys in
```
chris-wood, line 574:
```suggestion
```
The group seems implicit, so I think we can remove this.

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+- prk, pseudo-random key
+- creds, Credentials structure. Not used in this function, it only serves to comply with the API
+
+Output:
+- inner_env, nil value (serves to comply with the API)
+- pkc, the client's AKE public key
+
+Steps:
+1. _, pkc = DeriveKeyPair(prk)
+2. Output (nil, pkc)
```
chris-wood, line 586:
```suggestion
- client_public_key, the client's AKE public key

Steps:
1. _, client_public_key = DeriveKeyPair(prk)
2. Output (nil, client_public_key)
```

chris-wood: (Just renaming some things)

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
 ~~~
 
-The `custom_identifier` mode defines `SecretCredentials` and `CleartextCredentials` as follows:
+OPAQUE-3DH uses HashToScalar(seed) to derive the client's secret key in this mode.
+~~~
+DeriveKeyPair(prk)
```
chris-wood, line 591: I think we should move this to the configuration section, which discusses particular types of Groups for 3DH.

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
 We let `Npk` and `Nsk` denote the size of public and private keys, respectively,
 used in the AKE.
 
-# Offline Registration {#offline-phase}
+- Deterministic Key Pair Generation Function:
+  - DeriveKeyPair(seed): Derive a private and public key pair deterministically from the seed.
+  - HashToScalar(msg, dst): Hash arbitrary input to a scalar in the field of the AKE group as defined in {{I-D.irtf-cfrg-voprf}}, except with dst = "OPAQUE-HashToScalar".
```
chris-wood, line 403:
```suggestion
```

This isn't an API that's exposed outside of its use in DeriveKeyPair, right? If not, we can probably drop it from here, and describe it where we specify how to implement DeriveKeyPair. (In the configurations section?)

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+- inner_env, nil value (serves to comply with the API)
+- pkc, the client's AKE public key
+
+Steps:
+1. _, pkc = DeriveKeyPair(prk)
+2. Output (nil, pkc)
 ~~~
 
-The `custom_identifier` mode defines `SecretCredentials` and `CleartextCredentials` as follows:
+OPAQUE-3DH uses HashToScalar(seed) to derive the client's secret key in this mode.
```
chris-wood, line 589:
```suggestion
OPAQUE-3DH implements DeriveKeyPair(prk) as follows:
```

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+
+Input:
+- prk, pseudo-random key
+- inner_env, an InnerEnvelope structure. Not used in this function, it only serves to comply with the API
+
+Output:
+- secret_creds, SecretCredentials
+
+Steps:
+1. skc = KeyGenFromSeed(prk)
```
chris-wood, line 617:
```suggestion
1. (client_secret_key, _) = DeriveKeyPair(prk)
```

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+2. Create SecretCredentials secret_creds with skc
+3. Output secret_creds
+~~~
+
+### External Key mode {#external-mode}
 
-These credentials are embedded into the following `Envelope` structure with
-encryption and authentication.
+In this mode, the application provides the OPAQUE instantiation with the client's authentication material through the `Credentials` structure. An encryption key is generated from the hardened OPRF output and used to encrypt the client's private key, which is then stored encrypted in the `Envelope`. This allows for keys of any kind as long as they are compatible with the AKE mechanism (e.g. RSA or ECDSA in the SIGMA-I setup, not specified in this document). Applications can therefore use existing key pairs for OPAQUE-3DH.
```
chris-wood, line 624:
```suggestion
In this mode, the application provides the OPAQUE instantiation with the client's authentication material through the `Credentials` structure. An encryption key is generated from the hardened OPRF output and used to encrypt the client's private key, which is then stored encrypted in the `Envelope`. This allows for applications which prefer to generate keys using some existing method. 
```

draft-irtf-cfrg-opaque.md
```diff
@@ -398,145 +398,331 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+- secret_creds, SecretCredentials
+
+Steps:
+1. encrypted_creds = inner_env.encrypted_creds
+2. pseudorandom_pad = Expand(prk, "Pad", len(encrypted_creds))
+3. secret_creds = xor(encrypted_creds, pseudorandom_pad)
+4. Output secret_creds
+~~~
+
+## Credential File {#credential-file}
```
chris-wood, line 671: Flagging this as a section that changed in @kevinlewi's PR, too.

draft-irtf-cfrg-opaque.md
```diff
@@ -734,6 +896,7 @@ data
 struct {
     SerializedElement data;
     opaque server_public_key[Npk];
+    opaque client_public_key[Npk];
```
chris-wood, line 899: Why is this needed? In internal mode, clients can compute it from the seed. In external mode, clients can also presumably compute it from the private key. Is this for settings where clients can't compute the public key from a private key?

chris-wood (changes requested):
Second pass done -- looking good! The rebase will obviously change some things, but the factored out functions is _really_ nice. I have comments and questions about some of the semantic changes and presentation, but we can certainly work through them (here or online -- whichever you prefer!)

draft-irtf-cfrg-opaque.md
bytemare: `auth_tag` covers the entire envelope, indeed, but putting it before `inner_env` ensures it's always in the same position, regardless of the mode, and thus simplifies parsing. What do you think?

draft-irtf-cfrg-opaque.md
bytemare: Indeed. I don't know if we can presume that clients can recompute their public key from the private key alone.
As much as I dislike this, simply sending it back to the client covers all cases and simplifies the implementation.

But if you think we should assume public key regeneration possible from the private one I'm totally on board.

draft-irtf-cfrg-opaque.md
bytemare: Yes, it does! Well spotted.

draft-irtf-cfrg-opaque.md
bytemare: Indeed, I think this is a leftover of a drafting phase. In my implementation, I only use this field to pass the server public key to the server when responding on login. Thanks for spotting this !

draft-irtf-cfrg-opaque.md
bytemare: Also, deriving the public key from the private key pushes the `auth_tag` verification to be done after the secret is recovered, i.e. we have the secret open and in memory, before we can authenticate the envelope. But maybe I'm just being overly paranoid.

draft-irtf-cfrg-opaque.md
bytemare: And that I'm thinking of it, client_public_key isn't necessary either.

draft-irtf-cfrg-opaque.md
chris-wood: Yep, that makes sense. Let's leave this as-is.

draft-irtf-cfrg-opaque.md
chris-wood: Hmm... how does this simplify the implementation? Does the client need to check that the public key it receives is correct for the given private key?

draft-irtf-cfrg-opaque.md
bytemare: Fixed.

draft-irtf-cfrg-opaque.md
bytemare: Moved to the configurations sections, please have a look

draft-irtf-cfrg-opaque.md
bytemare: It simplifies the implementation in case you're using something else than POG and if you need extra data to generate your public key (e.g. for RSA). But I forgot we pinned this to 3DH only. I'll amend.

bytemare: Fixed. See https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/154/commits/ad77e8accc08c246dca858aeb9fbf503566ec506

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +397,342 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+} CleartextCredentials;
+~~~
 
-            (response, oprf_key) = CreateRegistrationResponse(request, server_public_key, credential_identifier, oprf_seed)
+The function CreateCleartextCredentials implements this logic:
 
-                               response
-                      <-------------------------
+~~~
+CreateCleartextCredentials(server_public_key, client_public_key, server_identity, client_identity)
```
bytemare, line 440: `CreateCleartextCredentials()`
The name might be misleading for the new reader.

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +397,342 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
-AKE keys (server_private_key, server_public_key) for use with the AKE, along with a Nh-byte oprf_seed. S can use
-the same pair of keys with multiple clients, and can opt to use multiple seeds (so long as they are
-kept consistent for each client). These steps can happen offline, i.e., before the registration phase.
+OPAQUE makes use of a structure `Envelope` to manage client credentials. This envelope holds information about its format and content for the client to obtain its authentication material.
 
-Once complete, the registration process proceeds as follows.
+OPAQUE allows applications to either provide custom client private and public keys for authentication or to generate them internally, making the application oblivious to the client's private key. Each public and private key value is an opaque byte string, specific to the AKE protocol in which OPAQUE is instantiated.
 
-## Credential Registration
+These two options are defined as the `Internal Key` and `External Key` modes.  See {{envelope-modes}} for their specifications.
```
afbase, line 412: >OPAQUE allows applications to either provide custom client private and public keys for authentication or to generate them internally, making the application oblivious to the client's private key. Each public and private key value is an opaque byte string, specific to the AKE protocol in which OPAQUE is instantiated.
>
>These two options are defined as the Internal Key and External Key modes. See {{envelope-modes}} for their specifications.

Could we maybe be even more explicit to state that the the `EnvelopeMode` is external when a separate application(s) provide custom client private and public keys for authentication? 

>or to generate them internally

And then a secondary statement which states the complimentary where the `EnvelopeMode` is internal when it is not the case when a separate application(s) provide custom client private and public keys for authentication. 

Apologies if I'm falling out of the sky into this PR.  I've been following OPAQUE and TLS OPAQUE loosely for a bit.

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +397,342 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
 
-                                record
-                      ------------------------->
+Output:
+- cleartext_credentials, a CleartextCredentials structure
+
+Steps:
+1. if server_public_key == nil || client_public_key == nil
+2.  raise NoPubKeyError
+3. if server_identity == nil
```
kevinlewi, line 454: We should define nil somewhere, or just turn this into pseudocode by saying "if server_identity not specified". Same for client_identity

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +397,342 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
-  address or normal account name.
-- server_identity: The server identity. This is typically a domain name, e.g., example.com.
-  See {{identities}} for information about this identity.
+mode : The `EnvelopeMode` used for the `Envelope`.
+
+nonce : A unique nonce of length `Nn` used to protect this Envelope.
+
+auth_tag : Authentication tag protecting the contents of the envelope, covering `EnvelopeMode`, envelope nonce, `InnerEnvelope` and `CleartextCredentials`.
+
+inner_env : A mode dependent `InnerEnvelope` specifying at least the `EnvelopeMode`.
```
kevinlewi, line 491: "specifying at least the `EnvelopeMode`."

That's no longer true, right?

draft-irtf-cfrg-opaque.md
kevinlewi: Yes, it was removed -- can you remove it here as well?

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +397,342 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
 
-Registration is executed between a client C and a
-server S. It is assumed S can identify C and the client can
-authenticate S during this registration phase. This is the only part
-in OPAQUE that requires an authenticated and confidential channel, either physical, out-of-band,
-PKI-based, etc. This section describes the registration flow, message encoding,
-and helper functions. Moreover, C has a key pair (client_private_key, client_public_key) for an AKE protocol
-which is suitable for use with OPAQUE; See {{online-phase}}. The private-public keys (client_private_key, client_public_key) may be randomly generated (using a cryptographically secure pseudorandom number generator) for the account or provided by the calling client.
-Clients MUST NOT use the same key pair (client_private_key, client_public_key) for two different accounts.
+Generally, nonces are of length `Nn` set to 32 bytes.
```
kevinlewi, line 404: Can we remove the word "generally"? Or are there exceptions to this rule? Also, is there ever a case where `Nn` is not equal to 32, and if so, can we just remove the use of this extra variable and hard-code 32 in place of `Nn`?

draft-irtf-cfrg-opaque.md
kevinlewi: Can you also perhaps add a reference in the doc to where a reader can find the definition(s) of `InnerEnvelope`? Because they haven't been defined yet and it can cause confusion

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +397,342 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+Steps:
+1. seed = Expand(prk, concat(nonce, "PrivateKey"), Nsk)
+2. client_secret_key, client_public_key = DeriveKeyPair(seed)
+3. Create SecretCredentials secret_creds with skc
+4. Output (secret_creds, client_public_key)
+~~~
+
+### External Key mode {#external-mode}
+
+With the internal key mode the `EnvelopeMode` value MUST be `external`, and the size `Ne` of the serialized `Envelope` is 1 + Nn + Nm + Nsk.
```
kevinlewi, line 623: internal key -> external key

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +397,342 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
-auth_tag
-: Authentication tag protecting the contents of the envelope,
-covering `InnerEnvelope` and `CleartextCredentials`.
+                               response
+                      <-------------------------
 
-The full procedure for constructing `Envelope` and `InnerEnvelope` from
-`SecretCredentials` and `CleartextCredentials` is described in {{finalize-request}}.
-Note that only `SecretCredentials` are stored in the `Envelope` (in encrypted form).
+ (record, export_key) = FinalizeRequest(password, creds, blind, response)
```
kevinlewi, line 724: The API for FinalizeRequest should be updated (first argument is the client_private_key now)

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +397,342 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
 We let `Npk` and `Nsk` denote the size of public and private keys, respectively,
 used in the AKE.
 
-# Offline Registration {#offline-phase}
+- Deterministic Key Generation Functions:
+  - RecoverPublicKey(private_key): Recover the public key related to the input private key.
+  - DeriveKeyPair(seed): Derive a private and public key pair deterministically from the seed.
```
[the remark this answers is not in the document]
chris-wood: Hmm, these don't correspond to the OPRF, though, but are for the prime-order group keys, right?

draft-irtf-cfrg-opaque.md
chris-wood, line 404:
```suggestion
Random nonces used in this protocol are of length `Nn = 32` bytes.
```

draft-irtf-cfrg-opaque.md
chris-wood: Alternatively, we can modify the parameter description to read:

```
- server_public_key, The encoded client public key for the AKE protocol, which is nil if not specified
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +397,342 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
-  See {{identities}} for information about this identity.
+mode : The `EnvelopeMode` used for the `Envelope`.
+
+nonce : A unique nonce of length `Nn` used to protect this Envelope.
+
+auth_tag : Authentication tag protecting the contents of the envelope, covering `EnvelopeMode`, envelope nonce, `InnerEnvelope` and `CleartextCredentials`.
+
+inner_env : A mode dependent `InnerEnvelope` specifying at least the `EnvelopeMode`.
+
+The size of the serialized envelope is denoted `Ne` and varies in function of the `InnerEnvelope` length used by the modes.
```
chris-wood, line 493:
```suggestion
The size of the serialized envelope is denoted `Ne` and varies based on the mode. The exact value for `Ne` is specified in {{internal-mode}} and {{external-mode}}.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +397,342 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+~~~
+CreateEnvelope(random_pwd, server_public_key, client_secret_key, creds)
+
+Parameters:
+- params, the MHF parameters established out of band
+- mode, the EnvelopeMode mode
+- Nn, the nonce length
+
+Input:
+- y, finalized OPRF output
```
chris-wood, line 506: I would prefer that the hardened output (prk) be passed to this function (this function already does a lot).

draft-irtf-cfrg-opaque.md
```diff
@@ -350,14 +349,19 @@ OPAKE was taken.
 OPAQUE relies on the following protocols and primitives:
 
 - Oblivious Pseudorandom Function (OPRF, {{I-D.irtf-cfrg-voprf}}, version -06):
+  - GenerateKeyPair(): Generate an OPRF private and public key. OPAQUE only requires an OPRF private key. We write `(oprf_key, _) = GenerateKeyPair()` to denote use
+    of this function for generating secret key `oprf_key` (and discarding the
+    corresponding public key).
```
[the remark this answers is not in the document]
chris-wood: This looks like the rebase dropped some stuff, which makes me somewhat worried. @kevinlewi is anything else missing?

draft-irtf-cfrg-opaque.md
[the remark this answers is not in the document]
chris-wood: Additionally, @kevinlewi's change introduced DeriveKeyPair on the OPRF API. We need a different function name for the prime-order group variant. Perhaps, `DeriveGroupKeyPair`?

draft-irtf-cfrg-opaque.md
```diff
@@ -1070,11 +1235,47 @@ such that the following conditions are met:
...
   interface in {{dependencies}}. Examples include Argon2 {{?I-D.irtf-cfrg-argon2}},
   scrypt {{?RFC7914}}, and PBKDF2 {{?RFC2898}} with fixed parameter choices.
-- EnvelopeMode value is as defined in {{credential-storage}}, and is one of
+- EnvelopeMode value is as defined in {{client-credential-storage}}, and is one of
   `base` or `custom_identifier`.
 - The Group mode identifies the group used in the OPAQUE-3DH AKE. This SHOULD
   match that of the OPRF. For example, if the OPRF is OPRF(ristretto255, SHA-512),
   then Group SHOULD be ristretto255.
+  
+To generate the private key in the `Internal` mode, OPAQUE-3DH implements `DeriveKeyPair(seed)` as follows:
```
chris-wood, line 1244: I know I was the one who suggested this be down in {{configurations}}, but now I'm second guessing that. Perhaps these should be kept in the external mode section? @kevinlewi, what do you think? (Also, is the generator G defined anywhere?)

chris-wood (changes requested):
Another round after my implementation and second pass -- nice work as usual!

draft-irtf-cfrg-opaque.md
bytemare: +1 for dropping generally

On the other hand, I'm less comfortable with hardcoding values. But if you both think we should revert to do it, I can live with that :)

bytemare: I just saw @chris-wood  suggestion !

draft-irtf-cfrg-opaque.md
bytemare: Perfect, much better 👌  thanks !

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +397,342 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+Parameters:
+- params, the MHF parameters established out of band
+- mode, the EnvelopeMode mode
+- Nn, the nonce length
+
+Input:
+- y, finalized OPRF output
+- server_public_key, the server's AKE public key
+- client_secret_key, the client's private key
+- creds, a Credentials structure
```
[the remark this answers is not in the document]
bytemare: i.e. dropping the need for the `Credentials` structure right? (I'm in favor)

draft-irtf-cfrg-opaque.md
bytemare: Yes, true, well spotted.

draft-irtf-cfrg-opaque.md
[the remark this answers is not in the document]
bytemare: Sorry !

hm, indeed, these are for the AKE.

1. Should I reintroduce Kevin's `- DeriveKeyPair(seed): Derive a private and public key pair deterministically from a seed.` into the OPRF section?
2. I suggest we call it `DeriveAkeKeyPair` to make it very explicit

draft-irtf-cfrg-opaque.md
bytemare: Done!

draft-irtf-cfrg-opaque.md
bytemare: external or internal?

draft-irtf-cfrg-opaque.md
bytemare: Hello 👋

The key can be provided by the same application or the wrapper protocol. A usage example is when the same application needs the private key for some other use (with the internal mode, the private key never leaves the AKE). So it's not just for use with external applications.

draft-irtf-cfrg-opaque.md
bytemare: No, the server public MUST NOT be nil, it's the client's key that can be nil

chris-wood: 🤦 duh, yes, sorry. This only applies to the client.

draft-irtf-cfrg-opaque.md
bytemare: It's only used in the internal mode, it makes total sense to define it there, yes.

draft-irtf-cfrg-opaque.md
bytemare: Does this answer your question? Happy to discuss if you don't agree :)

afbase: >A usage example is when the same application needs the private key for some other use (with the internal mode, the private key never leaves the AKE).

That is a lot clearer thank you @bytemare  !

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,343 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+- prk, pseudo-random key
+- nonce, a unique nonce of length `Nn`
+- client_private_key, empty value. Not used in this function, it only serves to comply with the API
+
+Output:
+- inner_env, nil value (serves to comply with the API)
+- client_public_key, the client's AKE public key
+
+Steps:
+1. seed = Expand(prk, concat(nonce, "PrivateKey"), Nsk)
```
kevinlewi, line 607: Why is the length of the seed that is passed as input to DeriveAkeKeyPair of length `Nsk`?

Shouldn't it be `Nh`?

cc: @chris-wood

bytemare: Yes, you're right !

draft-irtf-cfrg-opaque.md
bytemare: But you are right, this is a totally valid use case and we could include that example. Thanks!

bytemare: Do you think that the following phrase

> OPAQUE allows applications to either provide custom client private and public keys for authentication

covers your following use case not explicitly enough?

> when a separate application(s) provide custom client private and public keys for authentication

Even if provided by an external application, it is still the wrapper application that implements or integrates OPAQUE that will import the keys.

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,341 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+~~~
+CreateEnvelope(prk, server_public_key, client_private_key, creds)
+
+Parameter:
+- mode, the EnvelopeMode mode
+
+Input:
+- prk, pseudo-random key
+- server_public_key, The encoded server public key for the AKE protocol
+- client_private_key, The encoded client private key for the AKE protocol
```
bytemare, line 496: Add `This is nil in internal mode`

draft-irtf-cfrg-opaque.md
```diff
@@ -627,17 +814,13 @@ Steps:
 ### FinalizeRequest {#finalize-request}
 
 ~~~
-FinalizeRequest(password, creds, blind, response)
-
-Parameters:
-- params, the MHF parameters established out of band
-- mode, the InnerEnvelope mode
-- Nx, the output size of the Extract function
+FinalizeRequest(client_private_key, password, blind, response)
```
bytemare, line 817: In external mode, implementations can be free to add the public key here. If added, then the public key doesn't have to be recovered in BuildInnerEnvelope() and the function must be adapted accordingly.

cc @chris-wood

draft-irtf-cfrg-opaque.md
bytemare: ftr: reverting this change, as the amount of entropy in the seed should be no less than the amount of entropy in a private key

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
-authenticate S during this registration phase. This is the only part
-in OPAQUE that requires an authenticated and confidential channel, either physical, out-of-band,
-PKI-based, etc. This section describes the registration flow, message encoding,
-and helper functions. Moreover, C has a key pair (client_private_key, client_public_key) for an AKE protocol
-which is suitable for use with OPAQUE; See {{online-phase}}. The private-public keys (client_private_key, client_public_key) may be randomly generated (using a cryptographically secure pseudorandom number generator) for the account or provided by the calling client.
-Clients MUST NOT use the same key pair (client_private_key, client_public_key) for two different accounts.
+Random nonces used in this protocol are of length `Nn = 32` bytes.
 
-## Setup Phase {#setup-phase}
+We define `nil` as a lack of value.
```
chris-wood, line 400: Can we move this to the notation section above?

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
-kept consistent for each client). These steps can happen offline, i.e., before the registration phase.
+# Client Credential Storage {#client-credential-storage}
 
-Once complete, the registration process proceeds as follows.
+OPAQUE makes use of a structure `Envelope` to manage client credentials. This envelope holds information about its format and content for the client to obtain its authentication material.
 
-## Credential Registration
+OPAQUE allows applications to either provide custom client private and public keys for authentication or to generate them internally, making the application oblivious to the client's private key. Each public and private key value is an opaque byte string, specific to the AKE protocol in which OPAQUE is instantiated.
+
+These two options are defined as the `Internal Key` and `External Key` modes.  See {{envelope-modes}} for their specifications.
```
chris-wood, line 408: Can we call this "Internal" and "External", rather than "Internal Key" and "External Key"?

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+
+Credential material is used to pin key material to identities if desired. If no identity is given for a party, its value MUST default to its public key. The following types of credentials are hereby considered:
```
chris-wood, line 410:
```suggestion
Applications may pin key material to identities if desired. If no identity is given for a party, its value MUST default to its public key. The following types of application credential information are hereby considered:
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+
+- client_private_key: The encoded client private key for the AKE protocol.
+- client_public_key: The encoded client public key for the AKE protocol.
+- server_public_key: The encoded server public key for the AKE protocol.
+- client_identity: The client identity. This is an application-specific value, e.g., an e-mail address or normal account name. If not specified, it defaults to the client's public key.
+- server_identity: The server identity. This is typically a domain name, e.g., example.com.  If not specified, it defaults to the server's public key.
+  See {{identities}} for information about this identity.
+
+These values are used in the `CleartextCredentials` structure as follows:
```
chris-wood, line 419:
```suggestion
These credential values are used in the `CleartextCredentials` structure as follows:
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+struct {
+ opaque server_public_key[Npk];
+ opaque server_identity<1..2^16-1>;
+ opaque client_identity<1..2^16-1>;
+} CleartextCredentials;
+~~~
 
-                               request
-                      ------------------------->
+The function CreateCleartextCredentials implements this logic:
```
chris-wood, line 429:
```suggestion
The function CreateCleartextCredentials constructs a `CleartextCredentials` structure given application credential information.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+ opaque server_identity;
+ opaque client_identity;
+} Credentials;
+~~~
 
-Upon completion, S stores C's credentials for later use.
+## Envelope Structure {#envelope-structure}
 
-## Credential Storage {#credential-storage}
+A client `Envelope` is constructed independently from the `EnvelopeMode`, but offers an `InnerEnvelope` entry that modes can use as they see fit.
```
chris-wood, line 463:
```suggestion
A client `Envelope` is constructed independently from the `EnvelopeMode`, but offers an `InnerEnvelope` entry whose structure is determined by the mode. Future modes MAY introduce alternate `InnerEnvelope` contents.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+
+nonce : A unique nonce of length `Nn` used to protect this Envelope.
+
+auth_tag : Authentication tag protecting the contents of the envelope, covering `EnvelopeMode`, envelope nonce, `InnerEnvelope` and `CleartextCredentials`.
+
+inner_env : A mode dependent `InnerEnvelope` structure. See {{envelope-modes}} for its specifications.
+
+The size of the serialized envelope is denoted `Ne` and varies based on the mode. The exact value for `Ne` is specified in {{internal-mode}} and {{external-mode}}.
+
+The `CreateEnvelope` function allows a client to create its `Envelope` at registration.
```
chris-wood, line 485:
```suggestion
Clients create an `Envelope` at registration with the function `CreateEnvelope` defined below.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+Parameter:
+- mode, the EnvelopeMode mode
 
-These credentials are incorporated in the `SecretCredentials` and `CleartextCredentials` structs,
-depending on the mode set by the value of `EnvelopeMode`:
+Input:
+- prk, pseudo-random key
+- server_public_key, The encoded server public key for the AKE protocol
+- client_private_key, The encoded client private key for the AKE protocol. This is nil in the internal key mode
+- server_identity, The encoded server identity
```
chris-wood, line 500:
```suggestion
- server_identity, The optional encoded server identity
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+- client_identity, The encoded client identity
```
chris-wood, line 501:
```suggestion
- client_identity, The optional encoded client identity
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+3. export_key = Expand(prk, concat(envelope_nonce, "ExportKey"), Nh)
+4. masking_key = Expand(prk, "MaskingKey", Nh)
+5. inner_env, client_public_key = BuildInnerEnvelope(prk, envelope_nonce, client_private_key)
+6. cleartext_creds = CreateCleartextCredentials(server_public_key, client_public_key, server_identity, client_identity)
+7. auth_tag = MAC(auth_key, concat(mode, envelope_nonce, inner_env, cleartext_creds))
+8. Create Envelope envelope with (mode, envelope_nonce, auth_tag, inner_env)
+9. Output (envelope, client_public_key, masking_key, export_key)
+~~~
+
+The `RecoverEnvelope` function allows a client to recover its `Envelope` at authentication.
```
chris-wood, line 521:
```suggestion
Clients recover their `Envelope` during authentication with the `RecoverEnvelope` function defined below.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+4. cleartext_creds = CreateCleartextCredentials(server_public_key, client_public_key, creds.server_identity, creds.client_identity)
+5. expected_tag = MAC(auth_key, concat(mode, envelope_nonce, inner_env, cleartext_creds))
+6. If !ct_equal(envelope.auth_tag, expected_tag),
+     raise MacError
+7. Output (client_private_key, export_key)
+~~~
+
+## Envelope Modes {#envelope-modes}
+
+The `EnvelopeMode` specifies the structure and encoding of the corresponding `InnerEnvelope`. This document specifies the values of the two aformentionned modes:
```
chris-wood, line 552:
```suggestion
The `EnvelopeMode` specifies the structure and encoding of the corresponding `InnerEnvelope`. This document specifies the values of the two aformentioned modes:
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+ internal(1),
+ external(2),
+ (255)
 } EnvelopeMode;
 ~~~
 
-The `base` mode defines `SecretCredentials` and `CleartextCredentials` as follows:
+Each `EnvelopeMode` defines its own `InnerEnvelope` structure and must implement the following interface:
+- `inner_env, client_public_key = BuildInnerEnvelope(prk, nonce, client_private_key)`: build and return the mode's `InnerEnvelope` structure and the client's public key.
+- `client_private_key, client_public_key = RecoverKeys(prk, nonce, inner_env)`: recover and return the client's private and public keys for the AKE protocol.
```
chris-wood, line 564:
```suggestion
- `client_private_key, client_public_key = RecoverKeys(prk, nonce, inner_env)`: recover and return the client's private and public keys for the AKE protocol.

The implementation of this interface for both `internal` and `external` modes is in {{internal-mode}} and {{external-mode}}, respectively.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
+- public_key, the associated public key
+
+Steps:
+1. private_key = HashToScalar(seed, dst)
+2. public_key = private_key * G
+3. Output (private_key, public_key)
 ~~~
 
-The `custom_identifier` mode defines `SecretCredentials` and `CleartextCredentials` as follows:
+HashToScalar(msg, dst): Hash arbitrary input to a scalar in the field of the AKE group as defined in {{I-D.irtf-cfrg-voprf}}, except that dst = "OPAQUE-HashToScalar".
```
chris-wood, line 593:
```suggestion
HashToScalar(msg, dst) is as specified in {{I-D.irtf-cfrg-voprf}}, except that dst = "OPAQUE-HashToScalar".
```

draft-irtf-cfrg-opaque.md
```diff
@@ -393,152 +391,346 @@ This specification defines one particular AKE based on 3DH; see {{instantiations
...
 } InnerEnvelope;
+~~~
 
-struct {
-  InnerEnvelope inner_env;
-  opaque auth_tag[Nm];
-} Envelope;
+encrypted_creds : Encrypted client_private_key. Authentication of this field is ensured with the `AuthTag` in the envelope that covers this `InnerEnvelope`.
+
+If the implementation provides the `client_public_key` to `CreateEnvelope()`, then `BuildInnerEnvelope()` should spare the `RecoverPublicKey()` call as the key is already available.
```
chris-wood, line 645:
```suggestion
If the implementation provides the `client_public_key` to `CreateEnvelope()`, then `BuildInnerEnvelope()` MAY skip the `RecoverPublicKey()` call and use the public key provided.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1070,12 +1236,28 @@ such that the following conditions are met:
...
   interface in {{dependencies}}. Examples include Argon2 {{?I-D.irtf-cfrg-argon2}},
   scrypt {{?RFC7914}}, and PBKDF2 {{?RFC2898}} with fixed parameter choices.
-- EnvelopeMode value is as defined in {{credential-storage}}, and is one of
+- EnvelopeMode value is as defined in {{client-credential-storage}}, and is one of
   `base` or `custom_identifier`.
 - The Group mode identifies the group used in the OPAQUE-3DH AKE. This SHOULD
   match that of the OPRF. For example, if the OPRF is OPRF(ristretto255, SHA-512),
   then Group SHOULD be ristretto255.
 
+To recover a public key from a private key, OPAQUE-3DH implements `RecoverPublicKey(private_key)` as follows:
```
chris-wood, line 1245: I'm not sure how I feel about this function here. It seems sort of out of place. But we can keep it now and move it around later.

chris-wood (approved):

draft-irtf-cfrg-opaque.md
bytemare: I believe it's `aforementioned` ! :)

draft-irtf-cfrg-opaque.md
bytemare: `BuildInnerEnvelope()` doesn't need to use the public key

draft-irtf-cfrg-opaque.md
bytemare: Should we move this to the _dependencies_ section?

PR #155, opened 2021-03-03
Generalize all baked-in cryptographic algorithms.

### LINKED ISSUE #132: HKDF-Expand limits

chris-wood:
Starting as a draft for now, since this does technically add more dependencies, but I think it's more future proof. 

Once we agree on the contents here, I'll update the PoC to match. Note that this should not change any test vectors, but will in fact remove some!

Closes #132.


draft-irtf-cfrg-opaque.md
```diff
@@ -1020,36 +1027,35 @@ Specifically, a one-time-pad is derived from Ke2 and then used as input to XOR
...
-- OPRF(ristretto255, SHA-512)
-- OPRF(decaf448, SHA-512)
-- OPRF(P-256, SHA-256)
-- OPRF(P-384, SHA-512)
-- OPRF(P-521, SHA-512)
+An OPAQUE-3DH configuration is a tuple (OPRF, KDF, MAC, MHF, Group).
+The OPRF protocol is drawn from the "base mode" variant of {{I-D.irtf-cfrg-voprf}}.
+The KDF functions in this document use HKDF {{!RFC5869}}. The MAC functions in
+this document use HMAC {{!RFC2104}}. The Hash functions in this specification
+include SHA-256 and SHA-512. The MHF functions use Scrypt {{?RFC7914}}
```
stef, line 1040: why not argon2i?

chris-wood: One can certainly use argon2i, but this document (right now) only uses scrypt.

stef: that is clear. my question was why?

chris-wood: This is not a great answer, but.... we can't properly cite Argon2i as it's [not yet an RFC](https://datatracker.ietf.org/doc/draft-irtf-cfrg-argon2/). (The same is true of the OPRF draft, but that's a dependency we can't workaround.) 

We can certainly change this to argon2i, or whatever, provided there's a suitable reference and evidence that the new thing is more widely supported than scrypt. The purpose of this is to be a recommendation for applications should they not know what they're doing.

draft-irtf-cfrg-opaque.md
```diff
@@ -371,20 +371,32 @@ OPAQUE relies on the following protocols and primitives:
...
-
-- Authenticated Key Exchange (AKE, {{instantiations}}):
-  - Npk: The size of the public keys used for the key exchange protocol.
-  - Nsk: The size of the private keys used for the key exchange protocol.
+- Key Derivation Function (KDF):
+  - Extract(salt, ikm): Extract a pseudorandom key of fixed length `Nl` bytes from
+    input keying material `ikm` and an optional byte string `salt`.
+  - Expand(prk, info, L): Expand a pseudorandom key `prk` using optional string `info`
+    into `L` bytes of output keying material.
+  - Nl: The output size of the `Extract()` function in bytes.
```
bytemare, line 379: This isn't really important, but if we name `Nh` after the hash length and `Nm` for the mac length, we might as well call this one `Ne` for the `Extract()` output length.

draft-irtf-cfrg-opaque.md
```diff
@@ -371,20 +371,32 @@ OPAQUE relies on the following protocols and primitives:
...
+  - Nh: The output size of the `Hash()` function in bytes.
 
 - Memory Hard Function (MHF):
   - Harden(msg, params): Repeatedly apply a memory-hard function with parameters
     `params` to strengthen the input `msg` against offline dictionary attacks.
     This function also needs to satisfy collision resistance.
 
+- Authenticated Key Exchange (AKE, {{instantiations}}):
+  - Npk: The size of the public keys used for the key exchange protocol.
+  - Nsk: The size of the private keys used for the key exchange protocol.
```
bytemare, line 398: This seems a bit too generic.

The introductory phrase states
> OPAQUE relies on the following protocols and primitives:

So I understand this block is a generic spec not pinning 3DH. It's true we rely on an AKE as a 'Cryptographic Protocol and Algorithm Dependency' here, so maybe we should define the 4 functions to handle 3-way messages? (e.g. start, response, finalize client, finalize server)

draft-irtf-cfrg-opaque.md
```diff
@@ -371,20 +371,32 @@ OPAQUE relies on the following protocols and primitives:
...
+
```
bytemare, line 399: It can make sense to specify the necessary Group operations in 3DH.
A "3DH Group" (or so) block could outline the necessary functions (e.g. KeyGen() for the ephemeral keys (or _GetRandomScalar()_ and _ScalarBaseMult()_ ) and ScalarMult() ).

bytemare (changes requested):

draft-irtf-cfrg-opaque.md
```diff
@@ -1020,36 +1027,40 @@ Specifically, a one-time-pad is derived from Ke2 and then used as input to XOR
...
+  then Group SHOULD be ristretto255.
+
+Absent an application-specific profile, the following configurations are RECOMMENDED:
+
+- OPRF(ristretto255, SHA-512), HKDF-SHA-512, HMAC-SHA-512, SHA-512, Scrypt(32768,8,1), ristretto255
+- OPRF(P-256, SHA-256), HKDF-SHA-256, HMAC-SHA-256, SHA-256, Scrypt(32768,8,1), P-256
+
+Future configurations may specify different combinations of dependent algorithms,
+with the following consideration. The size of AKE public and private keys -- `Npk`
+and `Npk`, respectively -- must adhere to an output length limitations of the KDF
```
bytemare, line 1061: `Nsk`

bytemare (changes requested):

draft-irtf-cfrg-opaque.md
```diff
@@ -1020,36 +1027,40 @@ Specifically, a one-time-pad is derived from Ke2 and then used as input to XOR
...
-match that of the OPRF.
+An OPAQUE-3DH configuration is a tuple (OPRF, KDF, MAC, Hash, MHF, EnvelopeMode, Group)
+such that the following conditions are met:
+
+- The OPRF protocol uses the "base mode" variant of {{I-D.irtf-cfrg-voprf}} and implements
+  the interface in {{dependencies}}. Examples include OPRF(ristretto255, SHA-512) and
+  OPRF(P-256, SHA-256).
+- The KDF, MAC, and Hash functions implement the interfaces in {{dependencies}}.
+  Examples include HKDF {{!RFC5869}} for the KDF, HMAC {{!RFC2104}} for the MAC,
+  and SHA-256 and SHA-512 for the Hash functions.
```
bytemare, line 1044: Can we cite KMAC?

draft-irtf-cfrg-opaque.md
```diff
@@ -371,20 +371,32 @@ OPAQUE relies on the following protocols and primitives:
...
-- Cryptographic hash function:
-  - Hash(m): Compute the cryptographic hash of input message `m`. The type of the
-    hash is determined by the chosen OPRF group.
-  - Nh: The output size of the Hash function.
-
-- Authenticated Key Exchange (AKE, {{instantiations}}):
-  - Npk: The size of the public keys used for the key exchange protocol.
-  - Nsk: The size of the private keys used for the key exchange protocol.
+- Key Derivation Function (KDF):
+  - Extract(salt, ikm): Extract a pseudorandom key of fixed length `Nl` bytes from
```
kevinlewi, line 375: nit: you reference `Nl` here, but it is defined below -- I'd say just remove the reference

draft-irtf-cfrg-opaque.md
```diff
@@ -371,20 +371,32 @@ OPAQUE relies on the following protocols and primitives:
...
+    input keying material `ikm` and an optional byte string `salt`.
+  - Expand(prk, info, L): Expand a pseudorandom key `prk` using optional string `info`
+    into `L` bytes of output keying material.
+  - Nl: The output size of the `Extract()` function in bytes.
+
+- Message Authentication Code (MAC):
+  - MAC(key, msg): Compute a message authentication code over input `msg` with key
+    `key`, producing a fixed-length output of `Nm` bytes.
```
kevinlewi, line 383: nit: you reference `Nm` here, but it is defined below -- I'd say just remove the reference

draft-irtf-cfrg-opaque.md
```diff
@@ -371,20 +371,32 @@ OPAQUE relies on the following protocols and primitives:
...
+  - Nm: The output size of the `MAC()` function in bytes.
+
+- Hash Function:
+  - Hash(msg): Apply a cryptographic hash function to input `msg`, producing an
+    fixed-length digest of size `Nh` bytes.
```
kevinlewi, line 388: nit: you reference `Nh` here, but it is defined below -- I'd say just remove the reference

draft-irtf-cfrg-opaque.md
kevinlewi: @bytemare The problem is that there may be future AKEs that support 4-way (or more) messages, so I think even that API can't be specified here...

draft-irtf-cfrg-opaque.md
kevinlewi: This would be 3DH-specific, though. Also, presumably we would reuse the corresponding functions in the OPRF bullet point for 3DH.

kevinlewi (approved):
This looks good to me.

I believe we also need to adjust the test vectors to include settings for Nm, Nl, Nh, etc. to be all equal to 32, so if you could add that before committing, that would be great!

draft-irtf-cfrg-opaque.md
chris-wood: I considered that, but `Ne` conflicts with the VOPRF terminology for "size of Element", so I went with `Nl`.

draft-irtf-cfrg-opaque.md
chris-wood: This is borrowed text, but I agree. We can change this to a prime-order group, specifically for 3DH. And the appendix can discuss how this might change if it's needed for other AKEs.

draft-irtf-cfrg-opaque.md
chris-wood: See above -- I'll just say we assume a prime-order group. That will help your internal/external PR, too, since I'll make HashToScalar a part of that interface.

draft-irtf-cfrg-opaque.md
chris-wood, line 1061:
```suggestion
and `Nsk`, respectively -- must adhere to an output length limitations of the KDF
```

chris-wood: Good catch!

draft-irtf-cfrg-opaque.md
bytemare: What if the VOPRF terminology changes to `Np` for "size of Point"?

chris-wood: Point implies curve, which may not always be the case. :)

draft-irtf-cfrg-opaque.md
```diff
@@ -371,20 +371,32 @@ OPAQUE relies on the following protocols and primitives:
...
+    into `L` bytes of output keying material.
+  - Nl: The output size of the `Extract()` function in bytes.
+
+- Message Authentication Code (MAC):
+  - MAC(key, msg): Compute a message authentication code over input `msg` with key
+    `key`, producing a fixed-length output of `Nm` bytes.
+  - Nm: The output size of the `MAC()` function in bytes.
+
+- Hash Function:
+  - Hash(msg): Apply a cryptographic hash function to input `msg`, producing an
```
bytemare, line 387: For 3DH, should the "running hash" functions Write() and Sum() be specified ? Or should we not go into that much detail?

chris-wood: That's certainly too much detail, I think. Implementations can implement `Hash` with a streaming API or one-shot API.

draft-irtf-cfrg-opaque.md
chris-wood: This is notation borrowed from the HPKE draft, so I'm inclined to keep it. We can always remove in a followup change!

draft-irtf-cfrg-opaque.md
```diff
@@ -371,20 +371,32 @@ OPAQUE relies on the following protocols and primitives:
...
+    fixed-length digest of size `Nh` bytes.
+  - Nh: The output size of the `Hash()` function in bytes.
```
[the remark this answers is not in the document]
chris-wood: That's a good point, and it's not clear to me what the minimum bound on `Nh`. I suggest we file an issue and track this for a future change. :)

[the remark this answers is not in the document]
bytemare: #158

bytemare (approved):

draft-irtf-cfrg-opaque.md
bytemare: Resurrecting this, at the risk of being annoying, but what do you think about `Nx`?

chris-wood: Why that over `Nl`?

bytemare: isn't the second letter referring to the function name? if _e_ is already taken, _x_ refers more easily to "extract" than _l_

chris-wood: This is a pretty epic bike shed! I picked `l` for length. I do not care what we go with here.

bytemare:
All vectors are good ✔️

draft-irtf-cfrg-opaque.md
```diff
@@ -1009,9 +1017,9 @@ IKM = concat(epkU^eskS, epkU^skS, pkU^eskS)
 Clients and servers use keys Km2 and Km3 in computing KE2.mac and KE3.mac,
 respectively. These values are computed as follows:
 
-- KE2.mac = HMAC(Km2, Hash(concat(preamble, KE2.enc_server_info)), where
+- KE2.mac = MAC(Km2, Extract(concat(preamble, KE2.enc_server_info)), where
   preamble is as defined in {{derive-3dh}}.
-- KE3.mac = HMAC(Km3, Hash(concat(preamble, KE2.enc_server_info, KE2.mac)),
+- KE3.mac = MAC(Km3, Extract(concat(preamble, KE2.enc_server_info, KE2.mac)),
```
bytemare, line 1022: @chris-wood I missed this. Why are we moving from a running hash to an Extract()?

chris-wood: I missed this too -- I meant to revert this to `Hash`. Can you file an issue (and send PR)? :)

PR #126, opened 2021-01-30
Simplify envelope secret derivation

chris-wood:
Closes #120.

cc @hugokraw

Issue #120, opened 2021-01-21
Simplifying envelope construction


kevinlewi:
Here are a subset of the steps for envelope construction:

```
1. N = Unblind(blind, response.data)
2. y = Finalize(pwdU, N, "OPAQUE01")
3. rwdU = HKDF-Extract("rwdU", Harden(y, params))
4. Create SecretCredentials secret_creds with creds.skU
5. Create CleartextCredentials cleartext_creds with response.pkS
   and custom identifiers creds.idU and creds.idS if mode is customIdentifier
6. nonce = random(32)
7. pseudorandom_pad = HKDF-Expand(rwdU, concat(nonce, "Pad"), len(pt))
8. auth_key = HKDF-Expand(rwdU, concat(nonce, "AuthKey"), Nh)
9. export_key = HKDF-Expand(rwdU, concat(nonce, "ExportKey"), Nh)
...
```

Can we instead change this to:
```
3. nonce = random(32)
4. rwdU = HKDF-Extract(nonce, Harden(y, params))
5. Create SecretCredentials secret_creds with creds.skU
6. Create CleartextCredentials cleartext_creds with response.pkS
   and custom identifiers creds.idU and creds.idS if mode is customIdentifier
7. pseudorandom_pad = HKDF-Expand(rwdU, "Pad", len(pt))
8. auth_key = HKDF-Expand(rwdU, "AuthKey", Nh)
9. export_key = HKDF-Expand(rwdU, "ExportKey", Nh)
```
?

Additionally, we could consider bundling up steps 7-9 into a single HKDF-Expand call, and taking slices of that output for each of the individual variables.

bytemare:
Replacing the static salt with the nonce, does is change any security property except added entropy ?

kevinlewi:
Well, we are actually moving the incorporation of the nonce from steps 7-9 into step 3, so I don't think this is actually adding entropy. The only thing it accomplishes is removing the need for the static salt "rwdU".

bytemare:
If I'm not mistaken, until now rwdu was not derived from the nonce. What I mean is that the nonce has more entropy than the string "rwdu" :)

kevinlewi:
It is true that the rwdU will have more entropy now, but the only uses of `rwdU` in the current protocol are to be used to derive `pseudorandom_pad`, `auth_key`, and `export_key`, all of which incorporate the nonce into the HKDF. That's all I meant by saying that the overall entropy is the same.

[...]

PR #156, opened 2021-03-05
Addressing client enumeration section with protocol changes

### LINKED ISSUE #22: Lift user enumeration text to the OPAQUE AKE stage
We might consider standardizing the mechanism by which servers prevent user enumeration attacks.

kevinlewi:
- Introduces an `oprf_seed` parameter which must be generated during server setup and persisted across all registration/login requests (similar to `server_private_key`)
- Introduces a `cred_identifier` parameter -- this could be the same as `client_identity`, but does not need to be
- Derives `oprf_key` from `oprf_seed` and `cred_identifier` -- this results in altering `RegistrationResponse` and `CredentialResponse`
- Modifies the "Credential File" section to just store the RegistrationUpload object, while making clear that the `oprf_seed` and `server_private_key` need to be kept around too
- In `FinalizeRequest`, adds a new element `envelope_key` which must be stored as part of a `RegistrationUpload` object
- Reverts the changes described in #120 (PR #126), necessary in order for the decryption step involving `envelope_key` to work properly
- Introduces an `encryption_nonce` parameter, and steps for encrypting the envelope using `envelope_key`, in `CreateCredentialResponse`
- Adds a new `FakeCredentialResponse` function
- Introduces a new "N" parameter, `Nok` for representing the size of an OPRF key. Removes unused calls to `SerializeScalar` and `DeserializeScalar` from the VOPRF draft
- Removes the text corresponding to the "Client Enumeration" section

Closes #22.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -350,23 +350,18 @@ OPAKE was taken.
 OPAQUE relies on the following protocols and primitives:
 
 - Oblivious Pseudorandom Function (OPRF, {{I-D.irtf-cfrg-voprf}}, version -06):
-  - GenerateKeyPair(): Generate an OPRF private and public key. OPAQUE only requires
-    an OPRF private key. We write `(oprf_key, _) = GenerateKeyPair()` to denote use
-    of this function for generating secret key `oprf_key` (and discarding the
-    corresponding public key).
   - Blind(x): Convert input `x` into an element of the OPRF group, randomize it
     by some scalar `r`, producing `M`, and output (`r`, `M`).
   - Evaluate(k, M): Evaluate input element `M` using private key `k`, yielding
     output element `Z`.
   - Finalize(x, r, Z): Finalize the OPRF evaluation using input `x`,
     random scalar `r`, and evaluation output `Z`, yielding output `y`.
-  - SerializeScalar(s): Map a scalar `s` to a unique byte array `buf` of fixed
-    length.
-  - DeserializeScalar(buf): Map a byte array `buf` to a scalar `s`, or fail if
-    the input is not a valid byte representation of a scalar.
+  - DeriveKeyPair(seed): Derive a private and public key pair deterministically
+    from a seed.
   - SerializedElement: A serialized OPRF group element, a byte array of fixed
     length.
   - SerializedScalar: A serialized OPRF scalar, a byte array of fixed length.
+  - Nok: The size of an OPRF private key
 
 Note that we only need the base mode variant (as opposed to the verifiable mode
 variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}.
@@ -403,26 +398,32 @@ used in the AKE.
 Registration is executed between a client C and a
 server S. It is assumed S can identify C and the client can
 authenticate S during this registration phase. This is the only part
-in OPAQUE that requires an authenticated channel, either physical, out-of-band,
+in OPAQUE that requires an authenticated and confidential channel, either physical, out-of-band,
 PKI-based, etc. This section describes the registration flow, message encoding,
 and helper functions. Moreover, C has a key pair (client_private_key, client_public_key) for an AKE protocol
 which is suitable for use with OPAQUE; See {{online-phase}}. The private-public keys (client_private_key, client_public_key) may be randomly generated (using a cryptographically secure pseudorandom number generator) for the account or provided by the calling client.
 Clients MUST NOT use the same key pair (client_private_key, client_public_key) for two different accounts.
 
-To begin, C chooses its password, and S chooses its own pair of private-public
-AKE keys (server_private_key, server_public_key) for use with the AKE. S can use
-the same pair of keys with multiple clients. These steps can happen offline, i.e.,
-before the registration phase. Once complete, the registration process proceeds as follows:
+## Setup Phase {#setup-phase}
+
+In a setup phase, C chooses its password, and S chooses its own pair of private-public
+AKE keys (server_private_key, server_public_key) for use with the AKE, along with a Nh-byte oprf_seed. S can use
+the same pair of keys with multiple clients, and can opt to use multiple seeds (so long as they are
+kept consistent for each client). These steps can happen offline, i.e., before the registration phase.
+
+Once complete, the registration process proceeds as follows.
+
+## Credential Registration
 
 ~~~
- Client (password, creds)            Server (server_private_key, server_public_key)
+ Client (password, creds)            Server (server_private_key, server_public_key, credential_identifier, oprf_seed)
  --------------------------------------------------------------------
  (request, blind) = CreateRegistrationRequest(password)
 
                                request
                       ------------------------->
 
-            (response, oprf_key) = CreateRegistrationResponse(request, server_public_key)
+            (response, oprf_key) = CreateRegistrationResponse(request, server_public_key, credential_identifier, oprf_seed)
 
                                response
                       <-------------------------
@@ -438,8 +439,7 @@ before the registration phase. Once complete, the registration process proceeds 
 Both client and server MUST validate the other party's public key before use.
 See {{validation}} for more details.
 
-Upon completion, S stores C's credentials for later use. See {{credential-file}}
-for a recommended storage format.
+Upon completion, S stores C's credentials for later use.
 
 ## Credential Storage {#credential-storage}
 
@@ -538,6 +538,8 @@ are recommended to be stored in a `Credentials` object with the following named 
 - `client_identity`, an optional client identity (present only in the `custom_identifier` mode)
 - `server_identity`, an optional server identity (present only in the `custom_identifier` mode)
 
+Note that the total size of the Envelope is equal to Nsk + Nh + 33 bytes.
+
 ## Registration Messages
 
 ~~~
@@ -565,12 +567,16 @@ server_public_key
 ~~~
 struct {
     opaque client_public_key[Npk];
+    opaque masking_key[Nh];
     Envelope envelope;
 } RegistrationUpload;
 ~~~
 
 client_public_key
 : The client's encoded public key, corresponding to the private key `client_private_key`.
+
+masking_key
+: A key used by the server to preserve confidentiality of the envelope during login
 
 envelope
 : The client's `Envelope` structure.
@@ -598,18 +604,21 @@ Steps:
 ### CreateRegistrationResponse {#create-reg-response}
 
 ~~~
-CreateRegistrationResponse(request, server_public_key)
+CreateRegistrationResponse(request, server_public_key, credential_identifier, oprf_seed)
 
 Input:
 - request, a RegistrationRequest structure
 - server_public_key, the server's public key
+- credential_identifier, an identifier that uniquely represents the credential being
+  registered
+- oprf_seed, the server-side seed of Nh bytes used to generate an oprf_key
 
 Output:
 - response, a RegistrationResponse structure
 - oprf_key, the per-client OPRF key known only to the server
 
 Steps:
-1. (oprf_key, _) = GenerateKeyPair()
+1. (oprf_key, _) = DeriveKeyPair(Expand(oprf_seed, concat(credential_identifier, "OprfKey"), Nok))
 2. Z = Evaluate(oprf_key, request.data)
 3. Create RegistrationResponse response with (Z, server_public_key)
 4. Output (response, oprf_key)
@@ -638,21 +647,22 @@ Output:
 Steps:
 1. y = Finalize(password, blind, response.data)
 2. envelope_nonce = random(32)
-3. prk = Extract(envelope_nonce, Harden(y, params))
+3. prk = Extract("", Harden(y, params))
 4. Create SecretCredentials secret_creds with creds.client_private_key
 5. Create CleartextCredentials cleartext_creds with response.server_public_key
    and custom identifiers creds.client_identity and creds.server_identity if
    mode is custom_identifier
-6. pseudorandom_pad = Expand(prk, "Pad", len(secret_creds))
-7. auth_key = Expand(prk, "AuthKey", Nx)
-8. export_key = Expand(prk, "ExportKey", Nx)
-9. encrypted_creds = xor(secret_creds, pseudorandom_pad)
-10. Create InnerEnvelope inner_env
+6. pseudorandom_pad = Expand(prk, concat(envelope_nonce, "Pad"), len(secret_creds))
+7. auth_key = Expand(prk, concat(envelope_nonce, "AuthKey"), Nh)
+8. export_key = Expand(prk, concat(envelope_nonce, "ExportKey"), Nh)
+9. masking_key = Expand(prk, "MaskingKey", Nh)
+10. encrypted_creds = xor(secret_creds, pseudorandom_pad)
+11. Create InnerEnvelope inner_env
       with (mode, envelope_nonce, encrypted_creds)
-11. auth_tag = MAC(auth_key, concat(inner_env, cleartext_creds))
-12. Create Envelope envelope with (inner_env, auth_tag)
-13. Create RegistrationUpload record with (envelope, creds.client_public_key)
-14. Output (record, export_key)
+12. auth_tag = MAC(auth_key, concat(inner_env, cleartext_creds))
+13. Create Envelope envelope with (inner_env, auth_tag)
+14. Create RegistrationUpload record with (creds.client_public_key, masking_key, envelope)
+15. Output (record, export_key)
 ~~~
 
 The inputs to Extract and Expand are as specified in {{dependencies}}.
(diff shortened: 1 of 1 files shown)


hugokraw:
Hi Kevin, I can't say I did a thorough review but here are a few comments from a cursory reading. Nothing major.

- the name envelope_key is too generic, there are too many envelope-related keys, how about enum_masking_key
- till now, we only required the registration to go over authenticated channels, with the sending of the enum_masking_key we now need to explicitly assume authenticated and confidential channel (e.g., TLS)
- In the derivation of prk, prk = HKDF-Extract("prk", Harden(y, params)), replace "prk" (acting as salt) with 0 (as in HKDF-Extract(salt=0, IKM)). There is no real value-add in using a different string. A salt of zero is the indication that instead of an extractor you are assuming a random oracle. Domain separators (as "prk" can be thought of) are not part of HKDF-Extract but of HKDF-Expand (there is  good rationale behind that distinction)
- The operations under FakeCredentialResponse need to be defined as closely to the "real case", particularly to avoid identifying the fake executions via side channels, particularly timing attacks. So instead of a random() in encrypted_response = random(Npk + Nsk + Nh + 33), l you may want to specify that the server generates encrypted_response as a fresh encryption of 0's using some dummy encryption key (even this can result in some timing difference if the encryption key is fixed leading to some optimizations). 
- Rationale for the enumeration-prevention measures. While you have now eliminated the text that proposed (and asked feedback for)  these measures, some of this text needs to stay to provide rationale (otherwise this whole thing looks "suspicious). Moreover, the issue in the previous bullet (timing attacks) needs to be explained and implementations for which enumeration is a real threat need to take measures to do an as accurate as possible simulation of the real case.

kevinlewi:
See my most recent comment on #22 -- I believe we should not make the changes I proposed to the protocol after all.

hugokraw:
> why is confidentiality required?

I assume you are asking about the need for a confidential channel during registration. This is because the (secret) masking_key needs to be sent from client to server (before, we only needed an authenticated channel). In most cases, if you have an authenticated channel (e.g. TLS) then you have a confidential one too (in particular, you can always upgrade authenticated to confidential by sending a public key over the authenticated channel)

hugokraw:
It may be worth having language saying that other than envelope masking, which is part of the basic protocol and a must-to-have for interoperability, the rest of FakeCredentialResponse (such as the use of oprf_seed) becomes a pure server-side implementation decision.

hugokraw:
@chris-wood Could the IETF/CFRG impose on the server to implement  FakeCredentialResponse for standard-compliance even though it is not needed for interoperability? In principle, such compliance could be tested by simulating a user enumeration attacker.   But I doubt it is something IETF would mandate. What do you think?

draft-irtf-cfrg-opaque.md
```diff
@@ -585,18 +590,21 @@ Steps:
...
+  registered
+- oprf_seed, the server-side seed of Nh bytes used to generate an oprf_key
 
 Output:
 - response, a RegistrationResponse structure
 - oprf_key, the per-client OPRF key known only to the server
 
 Steps:
-1. (oprf_key, _) = GenerateKeyPair()
+1. oprf_key = HKDF-Expand(oprf_seed, concat("oprf key", cred_identifier), Nok)
```
chris-wood, line 607: I don't think this is a workable replacement. This basically type casts bytes to an OPRF private key. I think we should revert this, and instead use DeriveKeyPair(HKDF(...)).

chris-wood (commented):
> @chris-wood Could the IETF/CFRG impose on the server to implement FakeCredentialResponse for standard-compliance even though it is not needed for interoperability? In principle, such compliance could be tested by simulating a user enumeration attacker. But I doubt it is something IETF would mandate. What do you think?

Well, we can certainly say that servers MUST implement the protocol this way to achieve the desired security properties. (This is akin, in my view, to saying that servers MUST choose ephemeral TLS key shares for connections, but some may choose to reuse them.) Whether or not implementations follow the requirement is up to them. :-)

kevinlewi:
Rebased, added DeriveKeyPair reference, and added the following optional text to the end of the Client Enumeration section:

> Finally, note that server implementations may choose to forego the construction
of a simulated credential response message for an unregistered client if these client
enumeration attacks can be mitigated through other application-specific means.

draft-irtf-cfrg-opaque.md
kevinlewi: Addressed in the latest push

bytemare (changes requested):

chris-wood (changes requested):

draft-irtf-cfrg-opaque.md
```diff
@@ -403,26 +399,32 @@ used in the AKE.
...
 which is suitable for use with OPAQUE; See {{online-phase}}. The private-public keys (client_private_key, client_public_key) may be randomly generated (using a cryptographically secure pseudorandom number generator) for the account or provided by the calling client.
 Clients MUST NOT use the same key pair (client_private_key, client_public_key) for two different accounts.
 
-To begin, C chooses its password, and S chooses its own pair of private-public
-AKE keys (server_private_key, server_public_key) for use with the AKE. S can use
+## Setup Phase {#setup-phase}
+
+In a setup phase, C chooses its password, and S chooses its own pair of private-public
+AKE keys (server_private_key, server_public_key) for use with the AKE, along with a Nh-byte oprf_seed. S can use
 the same pair of keys with multiple clients. These steps can happen offline, i.e.,
```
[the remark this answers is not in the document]
hugokraw: It does not have to be one seed for all. For example, there could be a seed for all userid's that start with letter a, one for those that start with letter b, etc. But a userid should always use the same oprf seed. A change of seed can only happen at re-registration.

draft-irtf-cfrg-opaque.md
```diff
@@ -350,23 +350,19 @@ OPAKE was taken.
...
-  - SerializeScalar(s): Map a scalar `s` to a unique byte array `buf` of fixed
-    length.
-  - DeserializeScalar(buf): Map a byte array `buf` to a scalar `s`, or fail if
-    the input is not a valid byte representation of a scalar.
   - SerializedElement: A serialized OPRF group element, a byte array of fixed
     length.
   - SerializedScalar: A serialized OPRF scalar, a byte array of fixed length.
+  - Nok: The size of an OPRF private key
+
+- Deterministic Key Pair Generation Function:
```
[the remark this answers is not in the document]
kevinlewi: Yes, it is. Can you clarify what you mean by "move it up"? It is already under the OPRF section -- do you mean to move it up in front of `Blind(x)`, `Evaluate(k, M)`, etc?

draft-irtf-cfrg-opaque.md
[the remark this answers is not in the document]
kevinlewi: @chris-wood: How should we allow for the flexibility in this API, in terms of producing test vectors, etc. ? Does it suffice to just add a note here saying that multiple oprf_seeds can be used?

draft-irtf-cfrg-opaque.md
```diff
@@ -663,18 +675,8 @@ Upon completion of this function, the client MUST send `record` to the server.
...
-
-~~~
-struct {
-    SerializedScalar oprf_key;
-    opaque client_public_key[Npk];
-    Envelope envelope;
-} credential_file;
-~~~
+The server then directly stores the `record` object as the credential file for each client. Note that
+the values `oprf_seed` and `server_private_key` from the server's setup phase must also be persisted.
```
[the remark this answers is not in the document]
kevinlewi: I don't think that's necessary. In the same way we don't mention the storing of any user or server identifiers when CustomIdentifier mode is used, I don't think we need to mention that the credential file has to store the credential identifier. But we can add it too if you think it increases clarity

draft-irtf-cfrg-opaque.md
```diff
@@ -663,18 +675,8 @@ Upon completion of this function, the client MUST send `record` to the server.
...
-these values might be conveniently stored together.
-
-~~~
-struct {
-    SerializedScalar oprf_key;
-    opaque client_public_key[Npk];
-    Envelope envelope;
-} credential_file;
-~~~
+The server then directly stores the `record` object as the credential file for each client. Note that
```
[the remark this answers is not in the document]
kevinlewi: Folded!

draft-irtf-cfrg-opaque.md
```diff
@@ -733,8 +735,7 @@ data
 ~~~
 struct {
     SerializedElement data;
-    opaque server_public_key[Npk];
-    Envelope envelope;
+    opaque encrypted_response[Npk + Nsk + Nh + 33];
```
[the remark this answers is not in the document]
kevinlewi: It is. Thanks for the catch -- added!

draft-irtf-cfrg-opaque.md
```diff
@@ -598,18 +606,21 @@ Steps:
...
+  registered
+- oprf_seed, the server-side seed of Nh bytes used to generate an oprf_key
 
 Output:
 - response, a RegistrationResponse structure
 - oprf_key, the per-client OPRF key known only to the server
 
 Steps:
-1. (oprf_key, _) = GenerateKeyPair()
+1. (oprf_key, _) = DeriveKeyPair(Expand(oprf_seed, concat("oprf key", credential_identifier), Nok))
```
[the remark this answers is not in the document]
kevinlewi: Thanks, used "OprfKey" here instead, following the usual convention of CamelCase for acronyms that are longer than 2 letters.

kevinlewi:
Addressed above comments. Also changed the use of "CredentialResponsePad" literal string to use encryption_nonce instead.

kevinlewi:
Changed names:
- `encryption_nonce` -> `masking_nonce`
- `encrypted_response` -> `masked_response`

draft-irtf-cfrg-opaque.md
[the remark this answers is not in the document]
chris-wood: Yep! If it's an OPRF API, it should be in the same list as Blind() etc.

draft-irtf-cfrg-opaque.md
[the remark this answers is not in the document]
chris-wood: Yeah, I think noting that multiple seeds may be used is sufficient, but the test vectors should just use one seed for all users.

draft-irtf-cfrg-opaque.md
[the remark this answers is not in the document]
chris-wood: I'm mainly interested in reducing the number of parameters to our core functions, which is growing little by little. We can keep this as-is for now, but we should keep this possible change on the back burner.

draft-irtf-cfrg-opaque.md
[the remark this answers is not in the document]
kevinlewi: I guess what I am saying is, it is **already** in the same list as Blind() etc... unless I am misinterpreting what you are saying?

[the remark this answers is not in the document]
chris-wood: It's in a separate list, like the KDF and other functions below.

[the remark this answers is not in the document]
kevinlewi: OH I see what you mean -- I thought you were referring to the `Nok` parameter as needing to move up. But now I realize that you are referring to `DeriveKeyPair` being needed to move up. Got it! :) The comment line numbers this points to was a bit ambiguous :)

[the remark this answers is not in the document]
chris-wood: Ah, yep :) Sorry, I should have been more clear.

draft-irtf-cfrg-opaque.md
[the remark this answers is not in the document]
kevinlewi: Added a side note that multiple seeds can be used

draft-irtf-cfrg-opaque.md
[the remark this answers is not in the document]
kevinlewi: This should be resolved in the latest diff.

draft-irtf-cfrg-opaque.md
```diff
@@ -403,26 +398,32 @@ used in the AKE.
...
- Client (password, creds)            Server (server_private_key, server_public_key)
+ Client (password, creds)            Server (server_private_key, server_public_key, oprf_seed)
  --------------------------------------------------------------------
  (request, blind) = CreateRegistrationRequest(password)
 
                                request
                       ------------------------->
 
-            (response, oprf_key) = CreateRegistrationResponse(request, server_public_key)
+            (response, oprf_key) = CreateRegistrationResponse(request, server_public_key, credential_identifier, oprf_seed)
```
chris-wood, line 426: Where does `credential_identifier` come from? It should probably be included in the Server inputs on L419.

draft-irtf-cfrg-opaque.md
```diff
@@ -565,13 +567,17 @@ server_public_key
...
+    opaque masking_key[Nh];
     Envelope envelope;
 } RegistrationUpload;
 ~~~
 
 client_public_key
 : The client's encoded public key, corresponding to the private key `client_private_key`.
 
+masking_key
+: A key used by the server to preserve the confidentiality of the envelope during login
```
chris-wood, line 579:
```suggestion
: A key used by the server to preserve confidentiality of the envelope during login
```

draft-irtf-cfrg-opaque.md
```diff
@@ -638,21 +647,22 @@ Output:
 Steps:
 1. y = Finalize(password, blind, response.data)
 2. envelope_nonce = random(32)
-3. prk = Extract(envelope_nonce, Harden(y, params))
+3. prk = Extract(salt=0, Harden(y, params))
```
chris-wood, line 650:
```suggestion
3. prk = Extract("", Harden(y, params))
```

draft-irtf-cfrg-opaque.md
```diff
@@ -770,25 +767,48 @@ Steps:
...
 ~~~
-CreateCredentialResponse(request, server_public_key, credential_file)
+CreateCredentialResponse(request, server_public_key, credential_file, credential_identifier, oprf_seed)
 
 Input:
 - request, a CredentialRequest structure
 - server_public_key, the public key of the server
-- credential_file, the server's output from registration
-  (see {{credential-file}})
+- credential_file, an instance of RegistrationUpload which is the server's
```
chris-wood, line 784:
```suggestion
- record, an instance of RegistrationUpload which is the server's
```

chris-wood: (here and elsewhere, since we refer to the element of this type as `record` in the FinalizeRequest function.)

chris-wood (approved):
LGTM modulo some final nits!

draft-irtf-cfrg-opaque.md
kevinlewi: Added!

draft-irtf-cfrg-opaque.md
kevinlewi: Done!

draft-irtf-cfrg-opaque.md
kevinlewi: Cool, also changed the salt=0 instance to "" in the "OPAQUE-3DH Key Schedule" section

hugokraw: Actually, it is better to set the salt to 0. It does not have to be described as salt=0 but just 0 (instead of ""). In principle, any fixed salt value can do, but I prefer to see 0 being used as the default "fixed value". In particular, 0 is used in TLS 1.3 key derivation when a random value for salt is not available and here we want to stay as close as possible to TLS 1.3 key derivation.

chris-wood: For what it's worth, "" is the notation used in the TLS zero-length strings, which is what we want here, I think.

hugokraw: What is the meaning of 0 in the following TLS 1.3 figure. Is it the value 0 or a 0-length string?
             0
             |
             v
   PSK ->  HKDF-Extract = Early Secret

hugokraw: Actually I see the answer: "0" indicates a string of Hash.length bytes set to zero.

We should specify the same here.

chris-wood: That's a zero-length string (or NULL, or the equivalent of providing no salt at all), but I was referring to "" used elsewhere where context strings.

chris-wood: > Actually I see the answer: "0" indicates a string of Hash.length bytes set to zero.

I'm not sure this is best, since the KDF might not be HKDF (and therefore have an implied hash function underneath). I think nil/""/0/whatever are equivalent, and "" seems to be most clear (to me) and consistent with 8446.

hugokraw: Good point. If we are treating the KDF as a general Extract-then-Expand primitive then we cannot say "a string of Hash.length bytes set to zero" but we also cannot say the empty string as it may not make sense in some other KDF. If you want to be generic, then you should refer to this salt value as a "constant" (as opposed to a random one), maybe call it const_salt and set it to "a string of Hash.length bytes set to zero" in the case of HKDF.

PR #164, opened 2021-03-23
Implement client enumeration fix.

chris-wood:
Based on changes in #156.


poc/test_opaque_ake.sage
```diff
@@ -75,6 +76,7 @@ def test_core_protocol_serialization():
     skS, pkS = group.key_gen()
     skS_enc = group.serialize_scalar(skS)
     pkS_enc = group.serialize(pkS)
+    oprf_seed = random_bytes(config.oprf_suite.group.scalar_byte_length())
```
kevinlewi, line 79: I believe this should be `Nh` (length of the hash output) bytes long, because `oprf_seed` is fed into the KDF's expand function.

poc/test_opaque_ake.sage
```diff
@@ -132,24 +135,25 @@ def test_registration_and_authentication():
...
```
kevinlewi, line 138: same here

kevinlewi (changes requested):

kevinlewi (approved):
We can consider re-adding the N* parameters back in at some point

PR #168, opened 2021-04-01
Implement internal and external mode.

chris-wood:
Implements #154. Starting as a draft while we work through interop.


poc/opaque_core.sage
```diff
@@ -63,30 +63,64 @@ class OPAQUECore(object):
...
+        Nh = self.config.hash().digest_size
+        auth_key = self.config.kdf.expand(random_pwd, envelope_nonce + _as_bytes("AuthKey"), Nh)
+        export_key = self.config.kdf.expand(random_pwd, envelope_nonce + _as_bytes("ExportKey"), Nh)
+        masking_key = self.derive_masking_key(random_pwd)
 
-        pt = secret_creds.serialize()
-        auth_data = cleartext_creds.serialize()
+        inner_env, client_public_key = self.build_inner_envelope(random_pwd, envelope_nonce, creds.skU)
+        cleartext_creds = self.create_cleartext_credentials(server_public_key, client_public_key, creds.idS, creds.idU)
+        auth_tag = self.config.mac.mac(auth_key, inner_env.serialize() + cleartext_creds.serialize())
```
kevinlewi, line 110: auth_tag needs to be computed over the mode and the nonce as well, not just the inner_env and cleartext_creds

poc/test_opaque_ake.sage
```diff
@@ -180,103 +145,102 @@ def test_3DH():
...
+                # Protocol inputs
+                if idU:
+                    inputs["client_identity"] = to_hex(idU)
+                if idS:
+                    inputs["server_identity"] = to_hex(idS)
+                inputs["oprf_seed"] = to_hex(oprf_seed)
+                inputs["credential_identifier"] = to_hex(credential_identifier)
+                inputs["password"] = to_hex(pwdU)            
+                inputs["client_private_key"] = to_hex(skU_bytes)
+                inputs["client_public_key"] = to_hex(pkU_bytes)
```
kevinlewi, line 197: One problem: when these are derived from the randomized_pwd seed, these are no longer inputs -- they are intermediate values. In particular, I don't think these were updated as their corresponding values when derived from the randomized_pwd seed, and so they end up not matching...

poc/opaque_core.sage
```diff
@@ -63,30 +63,64 @@ class OPAQUECore(object):
...
+
+    def build_inner_envelope(self, random_pwd, envelope_nonce, client_private_key):
+        if self.config.mode == envelope_mode_internal:
+            Nsk = self.config.Nsk
+            seed = self.config.kdf.expand(random_pwd, envelope_nonce + _as_bytes("PrivateKey"), Nsk)
+            (_, client_public_key) = self.derive_group_key_pair(seed)
+            pk_bytes = self.config.group.serialize(client_public_key)
+            return (InnerEnvelope(), self.config.group.serialize(client_public_key))
+        if self.config.mode == envelope_mode_external:
+            secret_creds = SecretCredentials(client_private_key)
```
bytemare, line 86: We're dropping `SecretCredentials` and encrypt `client_private_key` directly

poc/opaque_core.sage
bytemare: Agreed, that's where the vectors don't match :)

bytemare (changes requested):

poc/test_opaque_ake.sage
bytemare: I confirm

bytemare:
Something quite peculiar: P384 and P512 in the internal mode only have a public key in the registration upload vectors that doesn't correspond to either the client public key vector (that is false anyway) and the recovered public key from the password.

In all other vectors, the public key in the registration upload is correct.

Issue #159, opened 2021-03-10
Test vector section should include configuration values for KDF, MAC, etc.


kevinlewi:
This includes specifying `Nm`, `Nx`, etc. values in addition to the KDF and MAC used.

Issue #157, opened 2021-03-05
Agree on a consistent format for all "magic string" constants


kevinlewi:
Here is a list of all of the constant strings currently specified in the protocol:

Strings relating to envelope construction:
- "Pad"
- "AuthKey"
- "ExportKey"

Strings relating to the AKE section:
- "3DH"
- "client mac"
- "handshake secret"
- "server mac"
- "handshake enc"
- "encryption pad"
- "session secret"
- "OPAQUE "

Strings coming from the VOPRF draft (out of scope for this draft, but here for reference):
- "VOPRF06-HashToGroup-"
- "VOPRF06-Finalize-"

Can we agree on a convention for all of these strings?

Here's my proposal: I say we change all of them (except for the VOPRF ones obviously) to look like "AuthKey", "HandshakeSecret", "SessionSecret", and to remove the space at the end of "OPAQUE ".

bytemare:
+1 for CamelCase and space deletion

chris-wood:
+1 as well!

bytemare:
For OPAQUE defined strings, how about this:

Envelope construction:
- "OPAQUE-HashToScalar"
- "Pad"
- "AuthKey"
- "ExportKey"
- "MaskingKey"
- "PrivateKey"
- "OprfKey"
- "CredentialResponsePad"

Strings relating to the AKE section:
- "3DH"
- "OPAQUE-" _(note the dash in lieu of the space, I believe this will lead to fewer implementation mishaps)_
- "ClientMac"
- "ServerMac"
- "SessionSecret"
- "EncryptionPad"
- "HandshakeEnc"
- "HandshakeSecret"

bytemare:
Currently, the list of HashTo* DST's used are:

| Designation | Value  | Used in | Defined in|
| ------------- | ------------- | ------------- | ------------- |
| ku h2s | nil | CreateRegistrationResponse(), CreateCredentialResponse() | / |
| skc h2s | "OPAQUE-HashToScalar" | DeriveAkeKeyPair() | OPAQUE  |
| password h2c | ? | VOPRF/Blind() | [VOPRF](https://github.com/cfrg/draft-irtf-cfrg-voprf) |
| masking key h2s | ? | VOPRF/DeriveKeyPair()| [VOPRF](https://github.com/cfrg/draft-irtf-cfrg-voprf) |
[...]

PR #174, opened 2021-04-15
Apply various editorial changes.

### LINKED ISSUE #165: Expand input name
We currently use `prk` as the input to `Expand()` calls for alignment with HKDF, but this deviates from the original paper which used `rwd`. The latter might be too easily confused with `pwd`, so how about we clarify and call it `randomized_pwd`? That would also be more in line with other naming conventions used in the doc, where we spell things out rather than use acronyms. 

@hugokraw, what do you think?

chris-wood:
Fixes #165, #159, and #157.


bytemare:
Vectors check out ✅

PR #175, opened 2021-04-16
Clarify hash function requirements.

### LINKED ISSUE #158: Hash output lengths
The new interfaces introduced in #155 decouple the hashing function from the underlying hash functions used in KDF and MAC. This opens the possibility to use an XOF for hashing. Per definition, they don't have a fixed size output, so a size for their use should be set.

I suggest we also set 'length dependencies' between the output lengths and desired key lengths. For example, the `Expand()` output that yields `auth_key` should be of sufficient size for `MAC()`, and so on.

cc @chris-wood @kevinlewi

chris-wood:
Closes #158.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -87,6 +87,14 @@ protocols"
     seriesinfo: WETICE
     date: 2000
 
+  FIPS202:
+    title: "SHA-3 Standard: Permutation-Based Hash and Extendable-Output Functions"
+    target: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.202.pdf
+    date: Aug, 2015
+    author:
+      -
+        org: National Institute of Standards and Technology (NIST)
+
   GMR06:
     title: "A method for making password-based key exchange resilient to server compromise"
     author:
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -1263,42 +1273,34 @@ such that the following conditions are met:
...
-and `Nsk`, respectively -- must adhere to an output length limitations of the KDF
-Expand function. If HKDF is used, this means Npk, Nsk <= 255 * Nx, where Nx is the
-output size of the underlying hash function. See {{RFC5869}} for details.
+with the following considerations:
+
+1. The size of AKE public and private keys -- `Npk` and `Nsk`, respectively -- must adhere
+to an output length limitations of the KDF Expand function. If HKDF is used, this means
+Npk, Nsk <= 255 * Nx, where Nx is the output size of the underlying hash function.
+See {{RFC5869}} for details.
+1. The output size of the Hash function SHOULD be long enough to produce a key for
```
chris-wood, line 1301: This is the main change.

draft-irtf-cfrg-opaque.md
```diff
@@ -1263,42 +1273,34 @@ such that the following conditions are met:
   OPRF(P-256, SHA-256).
 - The KDF, MAC, and Hash functions implement the interfaces in {{dependencies}}.
   Examples include HKDF {{RFC5869}} for the KDF, HMAC {{!RFC2104}} for the MAC,
-  and SHA-256 and SHA-512 for the Hash functions.
+  and SHA-256 and SHA-512 for the Hash functions. If an extensible output function
+  such as SHAKE128 {{FIPS202}} is used then the output length `Nh` MUST be chosen
+  to align with the target security level of the OPAQUE configuration. For example,
+  if the security parameter is 128-bits, then `Nh` SHOULD be at least 32 bytes.
```
chris-wood, line 1279: This is the second part of this change.

bytemare: What is `security parameter` referring to?

chris-wood, line 1279:
```suggestion
  if the target security parameter for the configuration is 128-bits, then `Nh` SHOULD be at least 32 bytes.
```

chris-wood: Clarified that it's the target security parameter for the OPAQUE configuration. We're not (yet) crisp about that, but we can always expand on that later.

bytemare (approved):

PR #182, opened 2021-04-20
Mode configuration

### LINKED ISSUE #180: Fix mode to application instanciation
Different envelope modes yield different inner-envelope lengths. This difference leaks when an application changes modes, and breaks a privacy property of the user enumeration mitigation.

Hence, if an application implements the client enumeration mitigation, it must stick to a single mode. This means it's part of the instance configuration and makes its specification inside the envelope redundant and unnecessary.

bytemare:
- the mode must be part of the application configuration and static (not change within the app's lifecycle)
- don't include envelope mode in the envelope itself
- don't include envelope mode in the envelope's auth_tag
- add internal mode as recommended parameter

closes #180


chris-wood (approved):
LGTM pending test vector updates!

PR #186, opened 2021-04-20
Specify envelope length in RegistrationUpload

### LINKED ISSUE #184: Specify envelope length in message structures
Currently, the envelope has not specified any length in the message specifications, and I believe that's because it's a newly available value. For consistency, we should add its serialized size `Ne`.

bytemare:
Closes #184

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -790,7 +790,7 @@ server_public_key
 struct {
     opaque client_public_key[Npk];
     opaque masking_key[Nh];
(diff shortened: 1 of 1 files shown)


kevinlewi (changes requested):
According to https://tools.ietf.org/html/rfc8446, Section 3, this change would mean that the field consists of `Ne` Envelope structs. I believe the current state of the draft is correct as it stands.

bytemare:
Thanks @kevinlewi ! I wasn't sure about that :)

PR #187, opened 2021-04-20
clarify export_key client exclusivity

### LINKED ISSUE #181: Specify that the export key is client-only
Reading the current text in the draft, one could believe that the `export_key` is produced by both client and server. This should be fixed to makes it clear it's only available to the client.

bytemare:
Closes #181


kevinlewi (approved):
LGTM

chris-wood (approved):

PR #201, opened 2021-04-30
Updates based on Hugo's review.

chris-wood:


draft-irtf-cfrg-opaque.md
```diff
@@ -736,7 +739,7 @@ as it is not used.
 
 ### External mode {#external-mode}
 
-This mode allows applications to import custom keys for the client. This
+This mode allows applications to import or generate keys for the client. This
```
bytemare, line 742: ` to import or generate`
While that's true, it may confuse the reader as it can easily be understood as "this mode generates keys for the client". But maybe that's just me :)

bytemare (approved):

draft-irtf-cfrg-opaque.md
```diff
@@ -17,6 +17,10 @@ author:
     name: Hugo Krawczyk
     organization: Algorand Foundation
     email: hugokraw@gmail.com
+ -
+    ins: D. Bourdrez
+    name: Daniel Bourdrez
+    email: d@bytema.re
```
bytemare, line 23:
```suggestion
    email: dan@bytema.re
```

bytemare (approved):

draft-irtf-cfrg-opaque.md
chris-wood: The point of this change was to to suggest that it doesn't prohibit from key generation to happen _within_ the registration flow. (That is, registration could call DeriveAuthKeyPair or GenerateAuthKeyPair, for example, depending on mode.) 🤷

bytemare: Ah yes, I see :) thanks!

PR #203, opened 2021-05-11
Identities are vector of bytes.

armfazh:


chris-wood (changes requested):
Good catch! I'd go with a slightly different solution.

armfazh:
Closed in favor of #209

PR #205, opened 2021-05-19
Fixes random generation for test vectors.

armfazh:
Also fixes the serialization of `CleartextCredentials`

```go
struct {
  uint8 server_public_key[Npk];
  uint8 server_identity<1..2^16-1>;
  uint8 client_identity<1..2^16-1>;
} CleartextCredentials;
```
File `poc/opaque_messages.sage`:

```patch 
     def serialize(self):
-        return self.pkS + encode_vector(self.idU) + encode_vector(self.idS)
+        return self.pkS + encode_vector(self.idS) + encode_vector(self.idU)
```


armfazh:
todo(armfazh): do not update voprf in this PR.

PR #209, opened 2021-05-20
Removing Credentials structure.

armfazh:
The credentials structure is not a message of the protocol, thus, it's upto the implementer its storage definition. 
Its content is now split in `server_identity` and `client_identity` variables.

Closes #203


bytemare:
Agreed, it has become useless

chris-wood (approved):
I'll update the Sage code in a followup PR.

Issue #169, opened 2021-04-01
3DH: highlight the usage of client_info and server_info


armfazh:
In the 3DH instantiation, the fields `client_info` and `server_info` are marked as optional and application-specific. 

```
client_info
Optional application-specific information to exchange during the protocol.

enc_server_info
Optional application-specific information to exchange during the protocol encrypted under key Ke2, defined below.
```

The request here is to highlight in the document what are the possible use cases of these fields, and to remark the relevance of having these fields vs not having them at all.

bytemare:
Some observations about these fields

I'm not aware of many significant use cases for these fields. I'm very interested in examples that would undoubtedly justify these fields are kept. I can come up with valid ones, but they are weird enough they don't seem to solve any existing problem.

An important point to consider The `server_info `has strong constraints.
_If it is used for client-specific values_, it may leak info about the client's existence and activity, harming the client enumeration mitigation effort to some degree. That means that it must be of constant length, whether the user exists or not. Keeping this coherent across and for all clients in time doesn't look to scale reasonably well.
This being considered, application-specific `server_info `might not even need encryption if we consider an attacker already has access to the application.

Then again, `server_info `in KE2 instead of being in a message post-OPAQUE must justify a reasonable practical, security, or performance point. This point might be very niche and application-specific as well, thus not justifying `server_info `to be kept in-band. Maybe extracting its concept and specifying it in the appendix would be an acceptable compromise.

Also, there's a very interesting thread in #74

afbase:
# Prior Art And Use Cases

If any language from the OPAQUE draft is referenced below, it is in reference to the language found in commit [e221eaf](https://github.com/cfrg/draft-irtf-cfrg-opaque/tree/e221eaf702c2f018ad73bdfd3e043abbf8706b1f).

>The request here is to highlight in the document what are the possible use cases of these fields, and to remark the relevance of having these fields vs not having them at all.

_Please_ critique my use cases with the critique, let's focus on whether the client_info and server_info is useful here.

I will follow up with prior art examples of where extra server information is provided in other authentication schemes in another comment (either Tuesday or Wednesday evening Pacific time)

## Use Case 1 - Load Balancer with Multiple Backend Servers

Suppose we have:
  1. a client, 
  1. loadbalancer (LB),
  1. N backend servers (ASCII Diagram -> N=3),
  1. The client has completed a registration flow

```
                        ┌──────┐
                        │Client│
                        └──┬───┘
                           │
[...]

PR #211, opened 2021-05-25
Remove client_info and server_info and add shared Context to transcript

bytemare:
Addresses #169 

tl;dr

- completely remove client_info and server_info
- replace "3DH" transcript prefix with "RFCXXXX" (to be replaced by the actual RFC id)
- introduce optional shared context info "info" in configuration, to be length-prefix encoded in the transcript: "RFCXXXX" + encoded(info) + ...
- POC uses "OPAQUE-POC" as shared info field


draft-irtf-cfrg-opaque.md
```diff
@@ -1330,11 +1316,58 @@ Derive-Secret(Secret, Label, Transcript-Hash) =
 
 Note that the Label parameter is not a NULL-terminated string.
 
+OPAQUE-3DH integrates configuration and additional information
```
chris-wood, line 1319: I would simplify this and just say that `context` is an arbitrary-length byte string, and it's encoded in the transcript preamble like so:

```
preamble = concat(I2OSP(len(context), 2), context, ...
```

We could give guidance for what this context should include, such as what you list below, but it _seems_ overly prescriptive to force this particular structure here.

chris-wood (changes requested):
Great start! I'd simplify the format and structure of `Context` somewhat. (Were you also planning on updating the code? If not, I can do it.)

draft-irtf-cfrg-opaque.md
```diff
@@ -1330,11 +1314,18 @@ Derive-Secret(Secret, Label, Transcript-Hash) =
 
 Note that the Label parameter is not a NULL-terminated string.
 
+OPAQUE-3DH can optionally include shared `context` information in the transcript,
+such as configuration parameters or application-specific info, e.g. "appXYZ-v1.2.3".
+This information MUST be encoded with a length-prefix when included in the preamble.
```
chris-wood, line 1319:
```suggestion
```

chris-wood (approved):

Issue #166, opened 2021-03-24
Adding test vectors for "faked" credential response?


kevinlewi:
In https://github.com/cfrg/draft-irtf-cfrg-opaque/issues/22, we added the option for a server to be able to issue a "fake" credential response in the absence of a password file for a registered user. The current draft also describes how this response should be constructed. Should we also include test vectors for this functionality?

bytemare:
closed by #214

PR #213, opened 2021-06-02
Implement fake credentials response

bytemare:
Addresses #166 

A global fake record is created, that should be used in place of a legitimate one:
- envelope data is all 0 bytes
- a fake client public key is generated to serve for the AKE
- a random fake masking key is used

Application consideration for #199 
- server must not evaluate oprf on nil or empty user id and must abort

next step:
- add more explanatory text to draft about the fake response


bytemare:
Closing in favor of #214

PR #214, opened 2021-06-02
Take #2 on fake credential test vectors

chris-wood:
A couple open questions:

1. Does this include everything one would need to verify that KE2 is generated correctly? That seems to be the only thing we actually need to test here. (I don't see much value in testing to make sure the client side, but maybe I'm missing something.)
2. How much work will it be to verify vectors in your respective stacks, @bytemare and @kevinlewi?


bytemare:
1. Since we need to fake a client record, and the envelope is set to 0s, there must be a fake client public key, atop the client identifier and KE2 in the vectors. I added a commit to patch that.
2. Now they work :)

poc/format_test_vectors.py
```diff
@@ -5,6 +5,7 @@
 import json
 
 config_keys = [
+    "Fake",
```
kevinlewi, line 8: Suggestion: change to "IsFake"?

chris-wood: Alternate proposal: move these vectors to a separate appendix in the doc, so they're not mixed with the real vectors. Would that make it more clear?

kevinlewi: I think that would actually! Great idea. (And sorry for the delayed response on this)

This means that we wouldn't need the registration portion at all.

In addition, could we still have the client_private_key as a parameter, so that implementations can generate the client_public_key associated with the fake record more easily? Thanks :)

chris-wood: Pushed! Please give it a shot.

kevinlewi:
I just pushed a commit which does the following:

- Introduced a new parameter, `fake_masking_key`, as part of the test vectors. Previously, we were just generating an all-zeros bytestring for this.
- Adds the inputs `fake_client_private_key`, `fake_client_public_key`, and `fake_masking_key` to all of the fake test vectors
- Removes the intermediates and unnecessary inputs/outputs that existed in fake test vectors, in particular moving KE1 to an input instead of an output, and ensuring that KE2 is the only output.
- Fixed a bug in the existing fake test vector generation case: the previous code was using the "real" client pkU to generate KE1 and KE2 instead of the fake one

The current commit produces test vectors which match my Rust implementation in opaque-ke.

kevinlewi:
Latest amendment to the commit renames `fake_client_private_key`, `fake_client_public_key`, and `fake_masking_key` back to just `client_private_key`, `client_public_key`, and `masking_key`

bytemare:
Looks good! Vectors pass 🎉 

Thank you for the modifications!

bytemare (approved):
Looks good! Vectors pass 🎉

Thank you for the modifications!

PR #218, opened 2021-06-07
Fix some errors. Clarify undefined things.

### LINKED ISSUE #217: Should `G` or `MacError` be clarified?
As far as I see, `MacError` is not properly defined. While I understand what it refers to, it might be good to define it for someone that might get confused.

On line 679, also, on the `DeriveAuthKeyPair`:

```
2. public_key = private_key * G
```
the value `G` is not defined elsewhere, which might be confusing for a casual reader.. should it be better defined?

chris-wood:
Closes #217.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1550,7 +1550,7 @@ Input:
 - ke3, a KE3 structure.
 
 Output:
-- session_key, the shared session secret if, and only if, KE3 is valid, nil otherwise.
+- session_key, the shared session secret if and only if KE3 is valid.
 
 Exceptions:
 - HandshakeError, when the handshake fails
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -377,14 +377,15 @@ variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}.
 
 OPAQUE additionally depends on an Authenticated Key Exchange (AKE) protocol.
 This specification defines one particular AKE based on 3DH; see {{ake-protocol}}.
+3DH assumes a prime-order group as described in {{I-D.irtf-cfrg-voprf, Section 2.1}}.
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -1545,10 +1552,13 @@ Input:
 Output:
 - session_key, the shared session secret if, and only if, KE3 is valid, nil otherwise.
```
chris-wood, line 1553:
```suggestion
- session_key, the shared session secret if and only if KE3 is valid.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1545,10 +1552,13 @@ Input:
...
+Exceptions:
+- HandshakeError, when the handshake fails
+
 Steps:
-1. if ct_equal(ke3.client_mac, state.expected_client_mac):
-2.    Output state.session_key
-3. Output nil
+1. if !ct_equal(ke3.client_mac, state.expected_client_mac):
+2.    raise HandshakeError
+3. Output state.session_key
```
chris-wood, line 1561: This change makes the client and server interfaces different: both either return something, or toss an exception.

kevinlewi (approved):
Great changes!

draft-irtf-cfrg-opaque.md
```diff
@@ -676,7 +680,7 @@ Output:
 
 Steps:
 1. private_key = HashToScalar(seed, dst="OPAQUE-HashToScalar")
-2. public_key = private_key * G
+2. public_key = ScalarBaseMult(private_key)
```
bytemare, line 683: `ScalarBaseMult` is not defined in this document, if I'm not mistaken

chris-wood: It's in the VOPRF document, Section 2.1 (see the citation added in this PR, which also covers where HashToScalar is introduced.)

PR #232, opened 2021-07-01
Clarify what's stored on the server during registration.

### LINKED ISSUE #221: Comment around encrypted credentials text
The description of OPAQUE reads:

```
OPAQUE consists of two stages: registration and authenticated key exchange. In the first stage, a client registers its password with the server and stores its encrypted credentials on the server.
```

I really like this description but it might not be completely accurate as, as far as I understand, in the Internal mode, no encrypted credentials are stored, as the the client's private and public keys are deterministically derived from the OPRF output. Might be good to note that if what I'm saying is indeed correct ;)

chris-wood:
Closes #221.

cc @claucece


kevinlewi (approved):

claucece:
awesome! thank you @chris-wood

PR #233, opened 2021-07-01
Clarify enumeration prevention limitations.

### LINKED ISSUE #210: Revisiting user enumeration
Currently, OPAQUE has support [1] for preventing user enumeration during online _authentication_. It does not, however, prevent an attacker from using the offline _registration_ flow. (This was somewhat touched in #22.) 

This raises an important question: is the complexity of the enumeration attack prevention and mitigation useful given that the registration flow may already be exploited to learn this same information? 

@kevinlewi points out that the registration flow may have different restrictions in terms of who can trigger it. For example:

- Registration may be done once during some offline “bootstrap” or configuration step, thereby removing any registration oracle entirely. 
- Registration flows may be rate limited or otherwise restricted to attackers, in contrast to the authentication flow which may not have such restrictions. 

In general, this seems to boil down the use case. I don’t have enough experience with asymmetric PAKE applications to know how registration typically works. I hope we can use this issue to track the question above. 

[1] Assuming that servers construct fake CredentialResponse values without obvious side channels when a user does not exist.

chris-wood:
Closes #210.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1848,7 +1848,7 @@ Hardening the output of the OPRF greatly increases the cost of an offline
 attack upon the compromise of the credential file at the server. Applications
 SHOULD select parameters that balance cost and complexity.
 
-## Preventing Client Enumeration {#preventing-client-enumeration}
+## Client Enumeration {#preventing-client-enumeration}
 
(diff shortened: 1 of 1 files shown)


bytemare (approved):

kevinlewi (approved):
Great!

PR #236, opened 2021-07-08
set state.ke1 in client Start, removed unused ke1 input

aldenml:
Hi, found this while implementing the protocol. You can always set the value outside, but I think adding this step `state.ke1 = ke1` inside `Start` is very convenient and aligned with the use of `state.ke1` inside `ClientFinalize`.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1433,7 +1433,7 @@ Steps:
 ~~~
 
 ~~~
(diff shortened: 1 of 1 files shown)


bytemare:
These are indeed leftovers, as the value has already been taken out in the declaration and elsewhere. Good catch!

bytemare (approved):

aldenml:
glad to have a contribution here

PR #272, opened 2021-09-22
Note that configuration agreement is necessary for the protocol to run.

### LINKED ISSUE #246: Configuration agreement
The configurations section is the first place where client/server agreement on configurations is noted. We should promote this to an earlier part of the draft, and make clear that configuration negotiation is _not_ part of the protocol. (That would be done by any wrapper protocol, for example.)

chris-wood:
Closes #246.


bytemare:
👍

bytemare (approved):

PR #273, opened 2021-09-22
Pin key derivation seed length to 32 bytes.

### LINKED ISSUE #249: Internal mode private key seed length
The seed used to derive the private key in internal mode is currently `Nsk`:

```
seed = Expand(randomized_pwd, concat(nonce, "PrivateKey"), Nsk)
```

If `Nsk` is large for the AKE, then this seed may be quite large, too. We probably don't need anything beyond 32 bytes of randomness here, regardless of the AKE protocol. Perhaps we drop `Nsk` here and just use a constant? Or introduce another constant instead?

The same question might also be asked of the `ikm` derivation:

```
ikm = Expand(oprf_seed, concat(credential_identifier, "OprfKey"), Nok)
```

cc @hugokraw

chris-wood:
Closes #249.


bytemare:
👍

bytemare (approved):

PR #274, opened 2021-09-22
Harden the harden function.

### LINKED ISSUE #245: MHF usage
OPAQUE currently assumes that the MHF output is pseudorandom, as it's currently used as the IKM input to the KDF:

```
randomized_pwd = Extract("", Harden(y, params))
```

We could fix this by requiring the MHF to output something that's pseudorandom, or by changing how we invoke Extract, e.g.:

```
randomized_pwd = Extract(Harden(y, params), y))
```

cc @cjpatton, @hugokraw

chris-wood:
Closes #245.


bytemare:
Looks good, vectors included 👍

bytemare (approved):

Issue #257, opened 2021-08-24
Discussion around having AKE keys in an HSM


bytemare:
#238 mentioned a use case of using X25519 AKE server private keys in an HSM.

We could mention this in the Security Considerations section.

PR #279, opened 2021-10-04
Add security consideration for HSM

bytemare:
Addresses #257

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1841,6 +1841,17 @@ NOT RECOMMENDED. Applications should move such checks to the client. Note that
 limited checks at the server are possible to implement, e.g., detecting repeated
 passwords.
 
+## AKE Private Key Storage
+
+The server's AKE private key can be securely stored in a Hardware Security Module (HSM) or
+similar. Upon compromise of the OPRF seed and client envelopes, this would prevent an
+attacker from using this data to mount a server spoofing attack.
+
+Supporting implementations need to consider allowing separate AKE and OPRF algorithms in the
+case the HSM does not support the ciphersuite used for the OPRF. Also, they must not require
+access to the AKE private key bytes, only support for generating shared keys for the AKE is
+necessary.
+
 # IANA Considerations
 
 This document makes no IANA requests.
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1845,7 +1845,7 @@ passwords.
 
 Server implementations of OPAQUE do not need access to the raw AKE private key. They only require
 the ability to compute shared secrets as specified in {{key-schedule-functions}}. Thus, applications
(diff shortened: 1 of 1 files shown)


daxpedda:
Looks perfect to me, I think the only thing left is to evaluate if adding test vectors for X25519 and X448 makes sense.

bytemare:
It makes sense to me, and I think @chris-wood was also in favor.

X25519 is widespread and has good support, X448 is not as much.

Including these vectors would be better done in another PR.

draft-irtf-cfrg-opaque.md
```diff
@@ -1841,6 +1841,17 @@ NOT RECOMMENDED. Applications should move such checks to the client. Note that
...
+## AKE Private Key Storage
+
+The server's AKE private key can be securely stored in a Hardware Security Module (HSM) or
+similar. Upon compromise of the OPRF seed and client envelopes, this would prevent an
+attacker from using this data to mount a server spoofing attack.
+
+Supporting implementations need to consider allowing separate AKE and OPRF algorithms in the
+case the HSM does not support the ciphersuite used for the OPRF. Also, they must not require
+access to the AKE private key bytes, only support for generating shared keys for the AKE is
+necessary.
```
chris-wood, line 1853:
```suggestion
Server implementations of OPAQUE do not need access to the raw AKE private key. They only require
the ability to compute shared secrets as specified in {{key-schedule-functions}}. Thus, applications
may store the server AKE private key can be stored in a Hardware Security Module (HSM) or
similar. Upon compromise of the OPRF seed and client envelopes, this would prevent an
attacker from using this data to mount a server spoofing attack. Supporting implementations 
need to consider allowing separate AKE and OPRF algorithms in cases where the HSM is 
incompatible with the OPRF algorithm.
```

chris-wood: I _think_ this makes the HSM use case and considerations a bit more clear.

chris-wood (approved):
LGTM pending suggestion!

chris-wood:
> Including these vectors would be better done in another PR.

What issue should we use to track this?

draft-irtf-cfrg-opaque.md
```diff
@@ -1841,6 +1841,16 @@ NOT RECOMMENDED. Applications should move such checks to the client. Note that
 limited checks at the server are possible to implement, e.g., detecting repeated
 passwords.
 
+## AKE Private Key Storage
+
+Server implementations of OPAQUE do not need access to the raw AKE private key. They only require
+the ability to compute shared secrets as specified in {{key-schedule-functions}}. Thus, applications
+may store the server AKE private key can be stored in a Hardware Security Module (HSM) or
```
bytemare, line 1848:
```suggestion
may store the server AKE private key in a Hardware Security Module (HSM) or
```

@chris-wood is this what you meant ? :)

chris-wood (approved):

PR #280, opened 2021-10-05
factorize 3DH key derivation description

bytemare:
Last PR to close #263

This factorizes common code between the client and server 3DH functions, which, I believe, helps readability.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1349,7 +1349,7 @@ transcript, such as configuration parameters or application-specific info, e.g.
 The OPAQUE-3DH key schedule requires a preamble, which is computed as follows.
 
 ~~~
-Preamble(client_identity, ke1, server_identity, inner_ke2)
+Preamble(client_identity, ke1, server_identity, ke2)
 
 Parameters:
 - context, optional shared context information.
@@ -1360,7 +1360,7 @@ Input:
 - ke1, a KE1 message structure.
 - server_identity, the optional encoded server identity, which is set
   to server_public_key if not specified.
-- inner_ke2, an inner_ke2 structure as defined in KE2.
+- ke2, a KE2 message structure.
 
 Output:
 - preamble, the protocol transcript with identities and messages.
@@ -1371,14 +1371,15 @@ Steps:
                      I2OSP(len(client_identity), 2), client_identity,
                      ke1,
                      I2OSP(len(server_identity), 2), server_identity,
-                     inner_ke2)
+                     ke2.credential_response,
+                     ke2.AuthResponse.server_nonce, ke2.AuthResponse.server_keyshare)
 2. Output preamble
 ~~~
 
 #### Shared Secret Derivation
 
-The OPAQUE-3DH shared secret derived during the key exchange protocol is computed
-using the following function.
+The OPAQUE-3DH shared secret derived during the key exchange protocol is
+computed using the following functions.
 
 ~~~
 TripleDHIKM(sk1, pk1, sk2, pk2, sk3, pk3)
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -1392,33 +1393,45 @@ Output:
 
 Steps:
 1. dh1 = SerializePublicKey(sk1 * pk1)
-2. dh2 = SerializePublicKey(sk2 * pk2)
-3. dh3 = SerializePublicKey(sk3 * pk3)
-4. Output concat(dh1, dh2, dh3)
+1. dh2 = SerializePublicKey(sk2 * pk2)
+1. dh3 = SerializePublicKey(sk3 * pk3)
+1. Output concat(dh1, dh2, dh3)
```
chris-wood, line 1398: Does this render correctly?

draft-irtf-cfrg-opaque.md
```diff
@@ -1392,33 +1393,45 @@ Output:
...
 
 Output:
-- Km2, a MAC authentication key.
-- Km3, a MAC authentication key.
 - session_key, the shared session secret.
+- server_mac, the server's MAC.
+- client_mac, the client's MAC.
 
 Steps:
+1. if role == "client",
```
chris-wood, line 1419: I don't really like this -- can we just pass `ikm` in as input?

chris-wood (changes requested):
I'm not a fan of this change as I don't think it improves readability.

draft-irtf-cfrg-opaque.md
bytemare: yeah, probably

draft-irtf-cfrg-opaque.md
bytemare: no, it doesn't, bad leftover. thanks!

bytemare:
Basically, I'm just centralizing the key derivation and mac calculation in a single function, so that the client and server functions don't repeat the same code. Also, there were some leftovers from other PRs that hadn't been fixed

draft-irtf-cfrg-opaque.md
```diff
@@ -1392,33 +1393,32 @@ Output:
 
 Steps:
 1. dh1 = SerializePublicKey(sk1 * pk1)
-2. dh2 = SerializePublicKey(sk2 * pk2)
+2 dh2 = SerializePublicKey(sk2 * pk2)
```
chris-wood, line 1396:
```suggestion
2. dh2 = SerializePublicKey(sk2 * pk2)
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1392,33 +1393,32 @@ Output:
...
 3. dh3 = SerializePublicKey(sk3 * pk3)
 4. Output concat(dh1, dh2, dh3)
 ~~~
 
-Using this shared secret, further keys used for encryption and authentication are
-computed using the following function.
-
 ~~~
-DeriveKeys(ikm, preamble)
+DeriveKeys(preamble, ikm)
```
chris-wood, line 1402: Why swap the parameter order?

chris-wood:
Is this just a bug fix PR now?

draft-irtf-cfrg-opaque.md
bytemare: going back and forth, it just ended up like that, it was not intended. switching back

draft-irtf-cfrg-opaque.md
```diff
@@ -1397,28 +1398,27 @@ Steps:
...
+- client_mac, the client's MAC.
 
 Steps:
 1. prk = Extract("", ikm)
 2. handshake_secret = Derive-Secret(prk, "HandshakeSecret", Hash(preamble))
 3. session_key = Derive-Secret(prk, "SessionKey", Hash(preamble))
 4. Km2 = Derive-Secret(handshake_secret, "ServerMAC", "")
 5. Km3 = Derive-Secret(handshake_secret, "ClientMAC", "")
-6. Output (Km2, Km3, session_key)
+6. server_mac = MAC(Km2, Hash(preamble))
```
chris-wood, line 1419: Why this change? (At some point we need to stop superfluous changes.)

bytemare: The Km keys are only needed for that purpose, so we might as well use them here directly. It avoids having these lines in `ClientFinalize()` and `Response()`, which saves a couple of lines there.

chris-wood: Well, the function name is called `DeriveKeys`, but now it's doing more than that. Should we name the function if we're going to keep this change?

Issue #281, opened 2021-10-05
Update OPRF to adapt to POPRF


bytemare:
The (V)OPRF spec has been updated to include the POPRF findings. We should integrate these changes since this spec depends on it.

_Update: POPRF merged with #282, but some questions remain_

- _Does it make sense to have client-side domain separation in both POPRF (info field) and the AKE (context string)?_
- _If not, what are the arguments to have it on one part or the other?_

gtank:
This adds an important security property for our (@celo-org's) OPAQUE use case. We can't strongly identify or authorize the users of our OPRF endpoint because it's part of an account recovery flow, so we rely on rate limiting client identities to prevent brute force attempts. As current specified OPAQUE seems to allow the following attack (which @nategraf and I spotted here and also in our current design with a different OPRF construction - it's a fundamental issue):

1. Attacker submits a fraudulent request for the credential file of a target identity. They receive the encrypted blob but fail to decrypt it. This preserves the main security guarantees, since it prevents attacker access to the stored credential or a valid session key, but it still gives them a guess at the password's OPRF evaluation. We assume rate limiting stops giving OPRF outputs in response to their password guesses at some point, but the attacker can retain the encrypted credential file of the target account.

2. Attacker then creates a sequence of new accounts and requests their own credential files, but continues guessing the target's password and applying the OPRF outputs to decryption attempts on the target file.

Basically, since the OPRF evaluation isn't domain-separated by identity, then the target's masking key can be derived from a successful guess by anyone. You could use different keys for different users (which is how some POPRFs work anyway) but a single-key POPRF would allow us to bind the client ID to the OPRF evaluation and enforce rate limits on a per-client basis without the complexity of additional keys.

chris-wood:
Overcome by #324.

PR #282, opened 2021-10-06
Integrate POPRF

bytemare:
Address #281

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -343,14 +343,19 @@ Version -07, with the following API and parameters:
 
 - Blind(x): Convert input `x` into an element of the OPRF group, randomize it
   by some scalar `r`, producing `M`, and output (`r`, `M`).
-- Evaluate(k, M): Evaluate input element `M` using private key `k`, yielding
-  output element `Z`.
-- Finalize(x, r, Z): Finalize the OPRF evaluation using input `x`,
-  random scalar `r`, and evaluation output `Z`, yielding output `y`.
+- Evaluate(k, M, info): Evaluate input element `M` using private key `k` and
+  public input (or metadata) `info`, yielding output element `Z`.
+- Finalize(x, r, Z, info): Finalize the OPRF evaluation using input `x`,
+  random scalar `r`, evaluation output `Z`, and public input (or metadata)
+  `info`, yielding output `y`.
 - DeriveKeyPair(seed): Derive a private and public key pair deterministically
   from a seed.
 - Noe: The size of a serialized OPRF group element.
 - Nok: The size of an OPRF private key.
+
+The public input `info` is set to the public information available to the client
+that's consistent across sessions, which is the the ciphersuite parameters and
+the client identity. 
 
 Note that we only need the base mode variant (as opposed to the verifiable mode
 variant) of the OPRF described in {{I-D.irtf-cfrg-voprf}}. The implementation of
@@ -855,13 +860,15 @@ Steps:
 ### CreateRegistrationResponse {#create-reg-response}
 
 ~~~
-CreateRegistrationResponse(request, server_public_key, credential_identifier, oprf_seed)
+CreateRegistrationResponse(request, server_public_key, credential_identifier,
+                            oprf_seed, client_identity)
 
 Input:
 - request, a RegistrationRequest structure.
 - server_public_key, the server's public key.
 - credential_identifier, an identifier that uniquely represents the credential.
 - oprf_seed, the seed of Nh bytes used by the server to generate an oprf_key.
+- client_identity, the encoded client identity.
 
 Output:
 - response, a RegistrationResponse structure.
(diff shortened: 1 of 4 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -887,20 +895,21 @@ Input:
...
 
 Output:
 - record, a RegistrationRecord structure.
 - export_key, an additional client key.
 
 Steps:
-1. y = Finalize(password, blind, response.data)
-2. randomized_pwd = Extract("", concat(y, Harden(y, params)))
-3. (envelope, client_public_key, masking_key, export_key) =
+1. info = client_identity
```
chris-wood, line 905: This temp variable seems unnecessary -- why not just pass `client_idenity` to Finalize? (Same for the instance below, and Evaluate call above)

poc/test_opaque_ake.sage
```diff
@@ -72,20 +73,21 @@ def test_core_protocol_serialization():
...
 
-    recovered_skU, recovered_pkS, recovered_export_key = core.recover_credentials(pwdU, cred_metadata, cred_response)
+    recovered_skU, recovered_pkS, recovered_export_key = core.recover_credentials(pwdU, cred_metadata, cred_response, info)
 
     # Check that recovered credentials match the registered credentials
     assert export_key == recovered_export_key
 
 def test_registration_and_authentication():
     idU = _as_bytes("Username")
+    info = _as_bytes("Username")
```
chris-wood, line 90: Why not pass `idU` directly?

chris-wood (changes requested):
LGTM overall, but I think there's some unnecessary temp variables used throughout. (Thanks for updating the VOPRF dependency, too!)

poc/test_opaque_ake.sage
bytemare: To test whether it solves the 'NoneType' problem

draft-irtf-cfrg-opaque.md
chris-wood: Using client_id directly seems simplest here. Can we just do that?

poc/test_opaque_ake.sage
chris-wood: So we'll revert this, yes?

gtank:
> i see a lot of changes, and see that the reason for this is POPRF, but the PR lacks any references to what POPRF is, why it is relevant to OPAQUE, and thus why this PR is beneficial. would you please provide the missing context? thank you.

Hi folks! If it's helpful, I can supply at least one other reason for the PORPF in OPAQUE from my work on a WhatsApp-style e2e backup recovery system: https://github.com/cfrg/draft-irtf-cfrg-opaque/issues/281#issuecomment-937386635

poc/test_opaque_ake.sage
bytemare: of course ! This PR is still in draft btw ;)

draft-irtf-cfrg-opaque.md
bytemare: Good point @gtank ! Thanks :)

draft-irtf-cfrg-opaque.md
```diff
@@ -1158,21 +1164,22 @@ Input:
...
 Output:
 - response, a CredentialResponse structure.
 
 Steps:
 1. seed = Expand(oprf_seed, concat(credential_identifier, "OprfKey"), Nok)
 2. (oprf_key, _) = DeriveKeyPair(seed)
-3. Z = Evaluate(oprf_key, request.data)
-4. masking_nonce = random(Nn)
-5. credential_response_pad = Expand(record.masking_key,
+4. Z = Evaluate(oprf_key, request.data, client_identity)
```
chris-wood, line 1175: Numbering is off here.

poc/opaque_core.sage
```diff
@@ -27,9 +27,11 @@ class OPAQUECore(object):
     def __init__(self, config):
         self.config = config
 
-    def derive_random_pwd(self, pwdU, response, blind):
+    def derive_random_pwd(self, pwdU, response, blind, info):
         oprf_context = SetupBaseClient(self.config.oprf_suite)
-        y = oprf_context.finalize(pwdU, blind, response.data, None, None)
+        if info == None:
```
chris-wood, line 32: Rather than do this check inline, can we move it to the calling code and assume that `info` is never `None`?

chris-wood (approved):
LGTM pending one editorial fix and a suggestion on the PoC. @kevinlewi, can you please check the vectors?

poc/opaque_core.sage
bytemare: info is None when the client identity is not given (e.g. like in the test vectors)

bytemare: oh, you mean assume it's never `None` just in this function ?

kevinlewi (commented):
```
 response = CreateRegistrationResponse(request,
                                       server_public_key,
                                       credential_identifier,
                                       oprf_seed)
```

(Lines 773-776) The CreateRegistrationResponse needs to include the additional parameter, client_identity.

bytemare:
Nice catch @kevinlewi !

bytemare:
@caw: as @kevinlewi pointed out in #281, it wouldn't hurt to allow apps to input a specific string to use for POPRF, and with some small benefits.

If you both don't mind, I'd prefer to land this PR in the current state and that (an)other PR(s) address(es) the following points:
- (_do_) augment the POPRF info to allow applications to put in arbitrary strings as a prefix
- (_decide_) what do we do if no client id is known by the client: is an empty value acceptable?
- (_research_) if we want tighter binding and would like to include configuration parameters atop the app-specific string, we need to find a format to represent this (or an identification number à la IANA), or state that it's part of the app-specific input info and that the OPAQUE spec doesn't specify a configuration format and serialization

bytemare:
Same for mine ✅

PR #287, opened 2021-11-17
Add list of paper<>spec design differences

chris-wood:
This list includes notable design differences from the original paper description of OPAQUE made to accommodate the specification. This is done to help track the delta between the protocol which has been analyzed and that which is being specified in CFRG.

cc @hugokraw

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1665,7 +1665,7 @@ protocols such as TLS.
 The specification as written here differs from the original cryptographic design in {{OPAQUE}}.
 The following list enumerates important differences:
 
-- Envelopes do not contain encrypted credentials. Instead, envelopes contain information used to derive private key material for the AKE.
+- Envelopes do not contain encrypted credentials. Instead, envelopes contain information used to derive client private key material for the AKE.
 - Per-user OPRF keys are derived from a client identity and cross-user seed. This is done to mitigate against client enumeration attacks. See {{preventing-client-enumeration}} for more details.
 - The protocol outputs an export key in addition to shared session key that can be used for application-specific purposes.
 - The protocol admits optional application-layer client and server identities. In the absence of these identities, client and server are authenticated against their public keys. 
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1667,7 +1667,7 @@ The following list enumerates important differences:
 
 - Envelopes do not contain encrypted credentials. Instead, envelopes contain information used to derive client private key material for the AKE.
 - Per-user OPRF keys are derived from a client identity and cross-user seed. This is done to mitigate against client enumeration attacks. See {{preventing-client-enumeration}} for more details.
-- The protocol outputs an export key in addition to shared session key that can be used for application-specific purposes.
+- The protocol outputs an export key for the client in addition to shared session key that can be used for application-specific purposes.
 - The protocol admits optional application-layer client and server identities. In the absence of these identities, client and server are authenticated against their public keys. 
 - The protocol admits application-specific context information configured out-of-band in the AKE transcript. This allows domain separation between different application uses of OPAQUE.
 - Servers use a separate identifier for computing OPRF evaluations and indexing into the password file storage, called the credential_identifier. This allows clients to change their application-layer identity (client_identity) without inducing server-side changes, e.g., by changing an email address associated with a given account. 
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1666,7 +1666,7 @@ The specification as written here differs from the original cryptographic design
 The following list enumerates important differences:
 
 - Envelopes do not contain encrypted credentials. Instead, envelopes contain information used to derive private key material for the AKE.
-- Per-user OPRF keys are derived from a client identity and cross-user seed. This was done to mitigate against client enumeration attacks. See {{preventing-client-enumeration}} for more details.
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -1658,6 +1658,20 @@ enables a variety of OPAQUE instantiations, from optimized
...
 protocols such as TLS.
 
+## Notable Design Differences 
+
+[[RFC EDITOR: Please delete this section before publication.]]
+
+The specification as written here differs from the original cryptographic design in {{OPAQUE}}.
+The following list enumerates important differences:
+
+- Envelopes do not contain encrypted credentials. Instead, envelopes contain information used to derive private key material for the AKE.
```
bytemare, line 1668:
```suggestion
- Envelopes do not contain encrypted credentials. Instead, envelopes contain information used to derive client private key material for the AKE.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1658,6 +1658,20 @@ enables a variety of OPAQUE instantiations, from optimized
...
+- Per-user OPRF keys are derived from a client identity and cross-user seed. This was done to mitigate against client enumeration attacks. See {{preventing-client-enumeration}} for more details.
```
bytemare, line 1669:
```suggestion
- Per-user OPRF keys are derived from a client identity and cross-user seed. This is done to mitigate against client enumeration attacks. See {{preventing-client-enumeration}} for more details.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1658,6 +1658,20 @@ enables a variety of OPAQUE instantiations, from optimized
...
+- The protocol outputs an export key in addition to shared session key that can be used for application-specific purposes.
```
bytemare, line 1670:
```suggestion
- The protocol outputs an export key for the client in addition to shared session key that can be used for application-specific purposes.
```

bytemare (changes requested):

chris-wood:
@bytemare suggestions applied.

hugokraw:
Two more changes relative to the JKX'18 paper:
- The masking of the envelope
- JKX has a simplified analysis where the server builds all elements of the envelope for which it learns the password. The spec has the much more secure instantiation (suggested in JKX'18 but not analyzed) where the client computes the envelope and the server never sees the password.

We will need to justify (to CFRG) why these changes do not void the analysis. This justification is quite straightforward in all but two of the changes: The derivation of the user's private key from the output of the OPRF (rather than transported in the envelope) and the way the envelope is built at registration time. Both changes are analyzed in a manuscript by Hesse-Jarecki-Krawczyk-Wood that we could (probably should) post publicly. I will need first to check with the HJKW authors. 

Btw, these two changes actually strengthen the protocol. Obviously not sharing the password with the server during registration as (over) simplified in JKX18 (needed to match GMR UC formalism) is a *good* thing. And deriving the private key from the OPRF output removes the issue of RKR encryption and equivocability that were needed in the JKX18 analysis.

chris-wood:
@hugokraw I added your two suggestions.

bytemare (approved):

bytemare:
Looks good to me :)

hugokraw:
Just to document how these changes relate to the original protocol as analyzed in JKX18, and how to extend the analysis to cover these changes:

> The specification as written here differs from the original cryptographic design in [OPAQUE]. The following list enumerates important differences:

> Clients construct envelope contents without revealing the password to the server, as described in Section 5, whereas the servers construct envelopes in [OPAQUE].

This change adds to the security of the protocol. [OPAQUE] considered the case where the envelope was constructed by the server for reasons of compatibility with previous UC modeling, no essential reason. A later paper, to be submitted shortly, analyzes the registration phase as specified here. This new paper can be posted if we are requested to do so.

> Envelopes do not contain encrypted credentials. Instead, envelopes contain information used to derive client private key material for the AKE.

This variant is also analyzed in the new paper referred to in the previous item.  This change improves the assumption behind the protocol by getting rid of equivocability and random key robustness (RKR) for the encryption function (the latter property is only required for authentication and achieved by HMAC).

> Envelopes are masked with a per-user masking key as a way of preventing client enumeration attacks. See Section 10.9 for more details.

This extension does not add to the security of OPAQUE as an aPAKE but only used to provide a defense against enumeration attacks. In the analysis, this key can be simulated as a (pseudo) random key.

> Per-user OPRF keys are derived from a client identity and cross-user seed as a mitigation against client enumeration attacks. See Section 10.9 for more details.

The analysis of OPAQUE assumes OPRF keys of different users are independently random or pseudorandom. Deriving these keys via a single PRF (i.e., with a single cross-user key) applied to users' identities satisfies this assumption.

> The protocol outputs an export key for the client in addition to shared session key that can be used for application-specific purposes.

This key is a pseudorandom value independent of other values in the protocol and have no influence in the security analysis (it can be simulated with a random output).

> The protocol admits optional application-layer client and server identities. In the absence of these identities, client and server are authenticated against their public keys.

Binding authentication to identities is part of the AKE part of OPAQUE. The type of identities and their semantics are application dependent and independent of the protocol analysis

> The protocol admits application-specific context information configured out-of-band in the AKE transcript. This allows domain separation between different application uses of OPAQUE.

This is a mechanism for the AKE component and is best practice as for domain separation between different applications of the protocol.

> Servers use a separate identifier for computing OPRF evaluations and indexing into the password file storage, called the credential_identifier. This allows clients to change their application-layer identity (client_identity) without inducing server-side changes, e.g., by changing an email address associated with a given account.

This mechanism is part of the derivation of OPRF keys via a single OPRF. As long as the derivation of different OPRF keys from a single OPRF have different PRF inputs, the protocol is secure. The choice of such inputs is up to the application.

PR #288, opened 2021-11-17
Update masking key description

chris-wood:
cc @hugokraw

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -830,7 +830,7 @@ client_public_key
 : The client's encoded public key, corresponding to the private key `client_private_key`.
 
 masking_key
-: A key used by the server to preserve confidentiality of the envelope during login
+: An encryption key used by the server to preserve confidentiality of the envelope during login
   to defend against client enumeration attacks.
 
 envelope
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -683,8 +683,8 @@ Input:
 Output:
 - envelope, the client's `Envelope` structure.
 - client_public_key, the client's AKE public key.
-- masking_key, a key used by the server to encrypt the
-  envelope during login.
+- masking_key, a key used by the server with the sole purpose 
+  of defending against client enumeration attacks.
 - export_key, an additional client key.
 
 Steps:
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -683,8 +683,8 @@ Input:
 Output:
 - envelope, the client's `Envelope` structure.
 - client_public_key, the client's AKE public key.
-- masking_key, a key used by the server to encrypt the
-  envelope during login.
+- masking_key, a key used by the server with the sole purpose
```
chris-wood, line 686:
```suggestion
- masking_key, an encryption key used by the server with the sole purpose 
```

draft-irtf-cfrg-opaque.md
```diff
@@ -830,7 +830,8 @@ client_public_key
 : The client's encoded public key, corresponding to the private key `client_private_key`.
 
 masking_key
-: A key used by the server to preserve confidentiality of the envelope during login.
+: A key used by the server to preserve confidentiality of the envelope during login
```
chris-wood, line 833:
```suggestion
: An encryption key used by the server to preserve confidentiality of the envelope during login
```

draft-irtf-cfrg-opaque.md
```diff
@@ -683,8 +683,8 @@ Input:
 Output:
 - envelope, the client's `Envelope` structure.
 - client_public_key, the client's AKE public key.
-- masking_key, a key used by the server to encrypt the
-  envelope during login.
+- masking_key, a key used by the server with the sole purpose 
+  of defending against client enumeration attacks.
```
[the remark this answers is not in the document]
chris-wood: I don't know if it's useful, but it probably doesn't hurt. Fix applied.

bytemare (approved):

hugokraw:
I see two places where the masking_key function is defined in English:
- masking_key, an encryption key used by the server with the sole purpose 
  of defending against client enumeration attacks.
- An encryption key used by the server to preserve confidentiality of the envelope during login
  to defend against client enumeration attacks.

Let's have only one definition (even if repeated in two places).
In any case use the first. The masking is not used for confidentiality of the envelope (this is not required). What the masking does is to re-randomize the envelope to avoid linkability of same-user logins and allow for fake envelopes as needed to prevent enumeration attacks.  This explanation can be added to the security considerations.

Also, _I would very much like_ to see in the Security Considerations a sentence saying that the use of the DeriveKeyPair function for computing the oprf_key is only needed as an artifact against enumeration attacks. The security of OPAQUE does not require this otherwise, i.e., it can choose OPRF keys for the different users as random independent (or pseudorandom)  keys. 

One reason to say this is that choosing OPRF keys via DeriveKeyPair is suboptimal for threshold implementations of the OPRF function (this reason need not be said in this RFC except if we want to say something about threshold implementations).

PR #300, opened 2022-01-06
Make pseudocode more readable

### LINKED ISSUE #244: Revisit pseudocode descriptions
Perhaps alignment with HPKE would improve things? One downside is that HPKE does not make parameter types clear, which is currently done in the OPAQUE pseudocode. While we're in here, we should unify capitalization for the parameters. Inconsistent examples include:

```
- envelope, the client's `Envelope` structure
...
- server_identity, The optional encoded server identity.
```

As an aside, this is probably a larger question for CFRG: how do we want to express pseudocode for implementers?

chris-wood:
Closes #244.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1463,7 +1463,7 @@ Output:
 
 def Start(credential_request):
   client_nonce = random(Nn)
-  client_secret, client_keyshare = GenerateAuthKeyPair()
+  (client_secret, client_keyshare) = GenerateAuthKeyPair()
   Create KE1 ke1 with (credential_request, client_nonce, client_keyshare)
   Populate state with ClientState(client_secret, ke1)
   return (ke1, client_secret)
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1531,12 +1531,12 @@ Output:
 def Response(server_identity, server_private_key, client_identity,
              client_public_key, ke1, credential_response):
   server_nonce = random(Nn)
-  server_secret, server_keyshare = GenerateAuthKeyPair()
+  server_private_keyshare, server_keyshare = GenerateAuthKeyPair()
   Create inner_ke2 ike2 with (ke1.credential_response, server_nonce, server_keyshare)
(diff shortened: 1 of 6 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -369,10 +369,10 @@ Output:
 - private_key, a private key.
 - public_key, the associated public key.
```
bytemare, line 370: #244 mentions the unification of capitalization in input argument comments, which is not in this PR. Is that planned for another PR?

chris-wood:
@kevinlewi, thoughts on readability?

draft-irtf-cfrg-opaque.md
kevinlewi: I think it is fine here, since we do not want to capitalize.

draft-irtf-cfrg-opaque.md
```diff
@@ -693,21 +695,25 @@ Input:
...
-8. auth_tag = MAC(auth_key, concat(envelope_nonce, cleartext_creds))
-9. Create Envelope envelope with (envelope_nonce, auth_tag)
-10. Output (envelope, client_public_key, masking_key, export_key)
+def Store(randomized_pwd, server_public_key, server_identity, client_identity):
+  envelope_nonce = random(Nn)
+  masking_key = Expand(randomized_pwd, "MaskingKey", Nh)
+  auth_key = Expand(randomized_pwd, concat(envelope_nonce, "AuthKey"), Nh)
+  export_key = Expand(randomized_pwd, concat(envelope_nonce, "ExportKey"), Nh)
+  seed = Expand(randomized_pwd, concat(envelope_nonce, "PrivateKey"), Nseed)
+  _, client_public_key = DeriveAuthKeyPair(seed)
```
kevinlewi, line 708: While we are here, I kindof feel like it would be cleaner to do:
`(_, client_public_key) = DeriveAuthKeyPair(seed)`

instead of:
`_, client_public_key = DeriveAuthKeyPair(seed)`

for everywhere where DeriveAuthKeyPair is called.

kevinlewi (approved):
Overall, really like the change! Just one comment about the tuple outputs.

bytemare (approved):

PR #286, opened 2021-11-12
Fixes several errors on input and output parameters.

armfazh:
After implementing OPAQUE I found several errors on the description of the algorithms.
This PR amends most of them, no changes on the algorithms nor test vectors.
Hope this PR helps to improve a bit the draft.

PR #303, opened 2022-01-09
fix dst in DeriveAuthKeyPair

aldenml:
This follows the code https://github.com/cfrg/draft-irtf-cfrg-opaque/blob/master/poc/opaque_core.sage#L65

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1317,7 +1317,7 @@ Output:
 - public_key, the associated public key.
 
 Steps:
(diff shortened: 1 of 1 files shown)


bytemare:
This is being addressed in #286

aldenml:
@bytemare thanks for the review and for pointing out this is addressed in #286, however, that PR is big and I have no idea if the merge ETA is short. Btw, this was an important element in trying to match the test vector for`KE1`. If you have veto power, I can close this immediately, otherwise, I hope you don't mind if I keep it open until @chris-wood have the time to look at it.

chris-wood:
I suspect #286 will be overcome by events, so let's land this now.

PR #306, opened 2022-01-09
Add Record encryption to Security Consideration

### LINKED ISSUE #240: Protect the client record with a KEM on registration
Currently, the client sends a masking key during registration for the server to obfuscate login messages and protect against client enumeration. This relatively sensitive and secret piece of data requires the full registration to be "secure" with integrity, authenticity, and confidentiality, as offered by TLS for example.

Integrity and authenticity cannot be subverted once the session is over and past. Confidentiality, however, is not ensured if encryption keys are backed up, shared, or leaked, and allows decryption after the session (e.g. proxy re-encryption, logging, legal compliance, etc.). The masking key is at risk of being revealed to a passive (even non-malicious) attacker after registration has happened.

This risk can be avoided by using a KEM to protect the last registration message from the client to the server. Implementations could leverage the fact the client gets the server public key in the registration response and use symmetric encryption using a basic KEM. We could either build it into OPAQUE (not that expensive, as it happens only once) or add it to the security considerations.

bytemare:
closes #240, #306, and #310

This suggests protecting the client's record during registration with HPKE.


chris-wood (changes requested):
I think we can simplify this greatly and simply note that application-layer encryption is one possible way to implement the registration channel, no? @kevinlewi, @hugokraw, what do you think?

draft-irtf-cfrg-opaque.md
```diff
@@ -1865,6 +1865,26 @@ response during registration as an oracle for whether a given client identity is
 registered. Applications should mitigate against this type of attack by rate
 limiting or otherwise restricting the registration flow.
 
+## Protecting the masking key during registration
+
+The user enumeration prevention method uses a symmetric encryption key generated
+by the client and sent to the server at registration in the clear. This relatively
+sensitive and secret piece of data requires the full registration to be secured
+with integrity, authenticity, and confidentiality, as offered by TLS, for example.
```
kevinlewi, line 1873: I would remove "authenticity" because it is redundant, or if it really needs to be there, at least replace with "authentication"

bytemare: There's no authentication at the registration phase. The wording could be misleading.

Regarding authenticity, it's with regard to server messages really coming from the server.

daxpedda:
If I understand correctly, #310 make this obsolete?

bytemare:
putting this as draft, awaiting feedback on #310

draft-irtf-cfrg-opaque.md
bytemare: Does this answer the question?

draft-irtf-cfrg-opaque.md
```diff
@@ -1865,6 +1865,26 @@ response during registration as an oracle for whether a given client identity is
...
+due to proxy re-encryption, logging, legal compliance). In transit, the masking
+key is subject to logging after TLS termination.
+
+In these cases, the masking key is at risk of being revealed to a passive (even
+non-malicious) attacker after registration has happened.
+
+This risk can be avoided by using a Key Encapsulation Mechanism (KEM) like HPKE
+{{?I-D.irtf-cfrg-hpke}}. Implementations would use the server's public key received
+in the registration response to encrypt the record sent in the last registration
+message from the client to the server.
```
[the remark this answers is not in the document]
bytemare: I'm not sure this really exposes the rationale of the recommandation

draft-irtf-cfrg-opaque.md
```diff
@@ -1865,6 +1865,39 @@ response during registration as an oracle for whether a given client identity is
 registered. Applications should mitigate against this type of attack by rate
 limiting or otherwise restricting the registration flow.
 
+## Protecting the Registration Masking Key
+
+The user enumeration prevention method described in this documents uses a symmetric
+encryption key generated by the client on regitration that is sent to the server in
+cleartext. This relatively sensitive and secret piece of data requires the full
+registration to be secured with integrity, authenticity, and confidentiality, as offered
+by TLS, for example.
```
chris-wood, line 1874:
```suggestion
encryption key generated by the client on registration that is sent to the server over
an authenticated channel, such as one provided by TLS {{RFC8446}}.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1865,6 +1865,39 @@ response during registration as an oracle for whether a given client identity is
...
+With TLS, integrity and authenticity can't be subverted once the session is over
+and past. Confidentiality, however, is not ensured if TLS encryption keys are
+backed up, shared, or leaked, and this allows decryption after the session (e.g.
+due to proxy re-encryption, logging, legal compliance). In transit, the masking
+key is subject to logging after TLS termination and therefore at risk of being
+revealed to a passive (even non-malicious) attacker after registration has happened.
```
chris-wood, line 1880:
```suggestion
In the event that this channel is compromised, this encryption key could be leaked
to an attacker.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1865,6 +1865,39 @@ response during registration as an oracle for whether a given client identity is
...
+
+Application layer encryption is one possible way to protect this secret in transit.
+
+Another possible method is to use a Key Encapsulation Mechanism (KEM) like HPKE
+{{?I-D.irtf-cfrg-hpke}} in the application layer. Clients would use the server's
+public key received in the registration response to encrypt the record sent in
+the last registration message from the client to the server.
```
chris-wood, line 1887:
```suggestion
One mitigation against this threat is to additionally encrypt the `RegistrationRecord` 
sent from client to server at the application layer using public key encryption, e.g., 
with HPKE {{?RFC9180}}.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1865,6 +1865,39 @@ response during registration as an oracle for whether a given client identity is
...
+the masking mechanism with a non-interactive DHKE to encrypt the envelope in transit
+during registration and the login phases. On registration, the client uses a
+server-provided public key (in the registration response) to encrypt its record to
+that key and send the result to the server. This adds confidentiality to the
+record as discussed above. On login, the server encrypts the envelope to the
+client's public key and sends that along in the credential response. The client
+recovers its private key (just like now) and can decrypt that ciphertext.
+
+The necessity of these mechanisms is realtive to the thread model the protocol
+is executed in. Their added complexity and are out of scope of this document.
```
chris-wood, line 1899:
```suggestion
```

I don't think we can keep this text since it would effectively yield a different protocol. Moreover, this is kind of redundant with the HPKE approach, so I don't think it adds much.

chris-wood (changes requested):
Mostly requesting simplifications.

draft-irtf-cfrg-opaque.md
bytemare: I'm not at all convinced that it yields a different protocol, as this only replaces the envelope encryption, itself an added component to the original OPAQUE protocol. I agree that the added complexity is out of scope for the document though, that's why we said we would mention it here.

cf the discussion in #310

chris-wood: I mean that the protocol mechanics change if this done rather than the existing encryption scheme. I agree that _functionally_ the protocol remains the same, but the implementation would be different. Hence, a different protocol.

bytemare: Would you accept a shorter, less detailed version of this paragraph stating this rationale? I think we should still mention that there _might_ be a problem for some threat models, and agree the countermeasures burden this spec

chris-wood: Sure, but isn't that already covered by the HPKE suggestion? I don't think any more detail is useful.

bytemare: Not really, the first addresses TLS compromission, and the second addresses the fact of having a client secret stored on the server. 

I updated the paragraph

chris-wood: I still don't get it. In both cases public key encryption is used. I think the latter case should be removed entirely.

bytemare: In the first case, the record holding the masking key is encrypted on registration. On login, the current symmetric method is used.

In the second case, the whole symmetric encryption is dropped: the user replaces the symmetric key with a public key on registration (therefore meaning risk number one is thwarted without encryption of the record), and the server encrypts the envelope with that provided key. Here the server does not need to store a secret value.

Both mechanisms use public-key encryption, yes, but only the second one addresses both risks. That's why I'm insisting on its value :)

draft-irtf-cfrg-opaque.md
```diff
@@ -1865,6 +1865,20 @@ response during registration as an oracle for whether a given client identity is
...
+The user enumeration prevention method described in this documents uses a symmetric
+encryption key generated by the client on registration that is sent to the server over
+an authenticated channel, such as one provided by TLS {{RFC8446}}. In the event that this
+channel is compromised, this encryption key could be leaked to an attacker.
+
+One mitigation against this threat is to additionally encrypt the `RegistrationRecord`
+sent from client to server at the application layer using public key encryption, e.g.,
+with HPKE {{?RFC9180}}.
+
+However, the threat models and the added complexity for this mechanim are out of scope
```
chris-wood, line 1879:
```suggestion
However, the details of this mechanism are out of scope
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1865,6 +1865,20 @@ response during registration as an oracle for whether a given client identity is
...
+
+The user enumeration prevention method described in this documents uses a symmetric
+encryption key generated by the client on registration that is sent to the server over
+an authenticated channel, such as one provided by TLS {{RFC8446}}. In the event that this
+channel is compromised, this encryption key could be leaked to an attacker.
+
+One mitigation against this threat is to additionally encrypt the `RegistrationRecord`
+sent from client to server at the application layer using public key encryption, e.g.,
+with HPKE {{?RFC9180}}.
+
```
chris-wood, line 1878:
```suggestion
```

chris-wood (approved):

Issue #312, opened 2022-01-12
Failure handling


daxpedda:
Currently it is not specified how to handle failures that aren't the fault of implementation, user or host (in the case of the usual authentication process).

The question on all the following points is:
- What should be done in the case of a failure:
  - Is it okay to just try again with a different input, for example in the case of randomness that caused the failure.
  - Should simply an error be emitted and the process started from scratch again.
- Should something be adjusted in the VOPRF spec to not produce an error.

This is a list of functions that can fail:
- [`CreateRegistrationRequest`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-07.html#section-5.2.1-1) can fail because it uses [`Blind`](https://www.ietf.org/archive/id/draft-irtf-cfrg-voprf-08.html#section-3.3.3.1-1), which can produce the identity point. I don't know if this is a problem or not, see cfrg/draft-irtf-cfrg-voprf#305. The easiest solution is to simply try again as the failure depends on RNG.

- The same applies to [`CreateRegistrationRequest`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-07.html#section-5.2.1-1).

- [`CreateRegistrationResponse`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-07.html#section-5.2.2-1) can fail because it uses [`Evaluate`](https://www.ietf.org/archive/id/draft-irtf-cfrg-voprf-08.html#section-3.3.1.1-1), which can simply fail with [`InverseError`](https://www.ietf.org/archive/id/draft-irtf-cfrg-voprf-08.html#section-5.1-2.3). In this case `CreateRegistrationResponse` can't be completed, unless a new `oprf_seed` is generated, or users receive an error and have to start from [`CreateRegistrationRequest`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-07.html#section-5.2.1-1) again.

- [`CreateCredentialResponse`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-07.html#section-6.3.2.2-3) can't fail, because otherwise `CreateRegistrationResponse` should have failed before, it might be helpful to mention that.

[...]

PR #314, opened 2022-01-16
Document errors

### LINKED ISSUE #256: Document possible errors
It would be very useful for implementers if we provide a minimal set of known errors that can happen during protocol execution, and what the suggested reaction to each error should be.

bytemare:
This documents some currently known errors.

closes #256 and #312


daxpedda:
As war as I can tell, this does not address #312. For example [`CreateRegistrationResponse`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-07.html#section-5.2.2-1) can fail with [`InverseError`](https://www.ietf.org/archive/id/draft-irtf-cfrg-voprf-08.html#section-5.1-2.3), which isn't mentioned here.

draft-irtf-cfrg-opaque.md
```diff
@@ -1646,6 +1646,62 @@ learn the client's password. This means that additional checks should be conside
 a production deployment of OPAQUE: for instance, ensuring that there is a
 server-authenticated channel over which OPAQUE registration and login is run.
 
+# API Considerations
+
+## Errors {#api-errors}
+
+This specification based on OPRF and 3DH only has a couple of fallible functions. For example,
+MACs exchanged during the AKE that might don't verify. Additionally, given the statically sized nature
```
daxpedda, line 1654:
```suggestion
MACs exchanged during the AKE that might not verify. Additionally, given the statically sized nature
```
Not exactly sure, english isn't my mother tongue.

bytemare: Mine neither :) but you're correct !

bytemare:
> As war as I can tell, this does not address #312. For example [`CreateRegistrationResponse`](https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-07.html#section-5.2.2-1) can fail with [`InverseError`](https://www.ietf.org/archive/id/draft-irtf-cfrg-voprf-08.html#section-5.1-2.3), which isn't mentioned here.

@daxpedda that's the POPRF error, correct?
https://github.com/cfrg/draft-irtf-cfrg-voprf/pull/304 brings back the "vanilla" OPRF, which OPAQUE will use once that PR is merged. The server won't be subject to that error anymore

daxpedda:
Thanks for the explanation @bytemare, looking forward to that then!

daxpedda:
Looking through cfrg/draft-irtf-cfrg-voprf#304, "vanilla" OPRF [`Blind`](https://cfrg.github.io/draft-irtf-cfrg-voprf/caw/all-three/draft-irtf-cfrg-voprf.html#section-3.3.1-2) can still produce a zero scalar. Is that something that should be addressed in the VOPRF spec or is that something that OPAQUE should handle? (I'm actually not sure if producing a zero scalar here would be a problem)

bytemare:
note to self: add the OPRF InvalidInputError

draft-irtf-cfrg-opaque.md
```diff
@@ -1646,6 +1646,63 @@ learn the client's password. This means that additional checks should be conside
 a production deployment of OPAQUE: for instance, ensuring that there is a
 server-authenticated channel over which OPAQUE registration and login is run.
 
+# API Considerations
+
+## Errors {#api-errors}
+
+This specification based on OPRF and 3DH only has a couple of fallible functions. For example,
+MACs exchanged during the AKE that might not verify. Additionally, given the statically sized nature
+of the protocol messages, clients and servers can easily detect errors in invalid messages or group elements.
```
chris-wood, line 1655:
```suggestion
MACs exchanged during the AKE might fail verification, yielding an error. Additionally, protocol messages 
sent between client and server might not match their expected size, yielding a different error.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1646,6 +1646,63 @@ learn the client's password. This means that additional checks should be conside
...
+|:-----------------------|:-------------|:-------------------------------------------------------------------------|
+| ErrInvalidMaskedLength | ClientFinish | The length of the masked response is not = point length + envelope size. |
+| ErrInvalidPKS          | ClientFinish | Invalid server public key in unmasked response.                          |
+| ErrEnvelopeInvalidMac  | ClientFinish | Invalid envelope authentication tag.                                     |
+| ErrAkeInvalidServerMac | ClientFinish | Invalid server AKE MAC.                                                  |
+
+
+### Server errors {#server-errors}
+
+The server returns a protocol error in the sole case of an invalid client MAC in the AKE.
```
chris-wood, line 1687:
```suggestion
The following table enumerates server errors that can occur, where they occur, and the reason for the error.
In this case, servers can only generate a single protocol error.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1646,6 +1646,63 @@ learn the client's password. This means that additional checks should be conside
...
+The server returns a protocol error in the sole case of an invalid client MAC in the AKE. 
+
+| Name                   | Stage         | Reason                  |
+|:-----------------------|:--------------|:------------------------|
+| ErrAkeInvalidClientMac | ServerFinish  | Invalid client AKE MAC. |
+
+### Application errors
+
+Implementations MUST verify certain input arguments and return an appropriate error. Note that
+these are not protocol errors, and are indicative.
```
chris-wood, line 1696:
```suggestion
these are not protocol errors, and are indicative. The following table enumerates server errors that can occur, where they occur, and the reason for the error.
```

chris-wood (changes requested):
I like this a lot! I'd like to move away from the Go-like naming convention, but otherwise, let's get this tidied up and landed.

bytemare:
@daxpedda We'd like to land this PR. Please don't hesitate if you think about other errors that we haven't covered here! :)

daxpedda (approved):
I was following this already, LGTM :+1:.

draft-irtf-cfrg-opaque.md
```diff
@@ -1665,6 +1663,64 @@ learn the client's password. This means that additional checks should be conside
...
+The following errors are documented for the OPRF-3DH composition in this specification and the recommended
+configurations, and do not cover potential errors in other configurations. Participants MUST abort the
+protocol when these errors occur.
+
+### Deserialization errors {#deserialization-errors}
+
+| Name                                 | Role   | Messages                  | Reason                                               |
+|:-------------------------------------|:-------|:--------------------------|:-----------------------------------------------------|
+| InvalidMessageLengthError            | Both   | All                       | The message length is invalid for the configuration. |
+| InvalidBlindedDataError              | Server | RegistrationRequest, Ke1  | Blinded data is an invalid point.                    |
```
chris-wood, line 1682:
```suggestion
| InvalidBlindedDataError              | Server | RegistrationRequest, Ke1  | Blinded data is an invalid element.                    |
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1665,6 +1663,64 @@ learn the client's password. This means that additional checks should be conside
...
+| InvalidEvaluatedDataError            | Client | RegistrationResponse, Ke2 | Invalid OPRF evaluation point.                       |
```
chris-wood, line 1683:
```suggestion
| InvalidEvaluatedDataError            | Client | RegistrationResponse, Ke2 | Invalid OPRF key derivation.                       |
```
Also, it doesn't look like this is actually raised in the spec?

draft-irtf-cfrg-opaque.md
chris-wood: Is this raised in the spec? I guess it's implicit, but we should probably note that otherwise.

draft-irtf-cfrg-opaque.md
```diff
@@ -1665,6 +1663,64 @@ learn the client's password. This means that additional checks should be conside
...
+| InvalidServerPublicKeyError          | Client | RegistrationResponse      | Invalid server public key.                           |
```
chris-wood, line 1684: What does "invalid public key" mean here? (Same comment as below)

draft-irtf-cfrg-opaque.md
```diff
@@ -1665,6 +1663,64 @@ learn the client's password. This means that additional checks should be conside
...
+
+### Client errors {#client-errors}
+
+The client can produce errors due to incorrect values in the messages it received. The following
+table enumerates some of these errors, where they occur, and the reason for the error.
+
+| Name                        | Stage                                                  | Reason                                                                   |
+|:----------------------------|:-------------------------------------------------------|:-------------------------------------------------------------------------|
+| InvalidInputError           | CreateRegistrationRequest,<br/>CreateCredentialRequest | OPRF input deterministically maps to the group identity element.         |
+| InvalidMaskedLengthError    | ClientFinish                                           | The length of the masked response is not = point length + envelope size. |
```
chris-wood, line 1697:
```suggestion
| InvalidMaskedLengthError    | ClientFinish                                           | Invalid masked response length. |
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1665,6 +1663,64 @@ learn the client's password. This means that additional checks should be conside
...
+| InvalidServerPublicKeyError | ClientFinish                                           | Invalid server public key in unmasked response.                          |
+| InvalidEnvelopeMACError     | ClientFinish                                           | Invalid envelope authentication tag.                                     |
```
chris-wood, line 1699:
```suggestion
| InvalidEnvelopeMACError     | ClientFinish                                           | Envelope authentication check failed.                                     |
```

(And likewise below?)

draft-irtf-cfrg-opaque.md
```diff
@@ -1491,7 +1489,7 @@ Output:
 - session_key, the shared session secret.
 
 Exceptions:
-- HandshakeError, when the handshake fails
+- InvalidServerAkeMACError, the handshake fails.
```
chris-wood, line 1492: Can we rename this to "HandshakeAuthenticationError", "ServerAuthenticationFailure," or similar? I don't know if we need to note "AKE" and "MAC".

draft-irtf-cfrg-opaque.md
```diff
@@ -1665,6 +1663,64 @@ learn the client's password. This means that additional checks should be conside
...
+### Application errors
+
+Implementations MUST verify certain input arguments and return an appropriate error. Note that
+these are not protocol errors, and are indicative. The following table enumerates server errors that can occur, where they occur, and the reason for the error.
+
+| Name                        | Role   | Stage      | Reason                                         |
+|:----------------------------|:-------|:-----------|:-----------------------------------------------|
+| InvalidServerSecretKeyError | Server | LoginInit  | Invalid server secret key.                     |
+| InvalidServerPublicKeyError | Server | LoginInit  | Invalid server public key.                     |
+| InvalidOPRFSeedLengthError  | Server | LoginInit  | Invalid OPRF seed length.                      |
```
chris-wood, line 1721: Why would this be thrown from the protocol, rather than enforced as a pre-condition to LoginInit? (Also, what is the LoginInit stage? It's not listed in the doc.)

draft-irtf-cfrg-opaque.md
```diff
@@ -1665,6 +1663,64 @@ learn the client's password. This means that additional checks should be conside
...
+| InvalidServerEphemeralPublicKeyError | Client | Ke2                       | Invalid ephemeral server public key.                 |
+
+### Client errors {#client-errors}
+
+The client can produce errors due to incorrect values in the messages it received. The following
+table enumerates some of these errors, where they occur, and the reason for the error.
+
+| Name                        | Stage                                                  | Reason                                                                   |
+|:----------------------------|:-------------------------------------------------------|:-------------------------------------------------------------------------|
+| InvalidInputError           | CreateRegistrationRequest,<br/>CreateCredentialRequest | OPRF input deterministically maps to the group identity element.         |
```
chris-wood, line 1696:
```suggestion
| InvalidInputError           | CreateRegistrationRequest, CreateCredentialRequest | OPRF input deterministically maps to the group identity element.         |
```

chris-wood (changes requested):
I think we need to sharpen some of the descriptions here, update some names, and mark how noted errors can be thrown in the rest of the doc.

draft-irtf-cfrg-opaque.md
bytemare: Isn't that more telling?
I fear that ServerAuthenticationFailure is too vague and doesn't say what the actual error is. e.g. it could be mistaken with the server sending the wrong public key (or whatever - this would result in the same error). I like the precision of this wording to pinpoint what went wrong.

draft-irtf-cfrg-opaque.md
bytemare: What do you mean?

bytemare: Yes, it's implicit. How do you suggest we should specify that?

bytemare: Add a second table with the implicit exceptions that come from other specs?

draft-irtf-cfrg-opaque.md
bytemare: Yes, I wanted to discuss this in the PR-draft stage. As stated, implementations must observe these conditions, and can raise these errors, since these conditions are necessary for the protocol to execute correctly

bytemare: I suppose LoginInit comes from my implementation, my bad! Amending right now

draft-irtf-cfrg-opaque.md
bytemare: Yes, many are implicit.

What about adding a dedicated table for implicit errors coming from other specs?

chris-wood: Can't we just add these alongside the existing errors? I think it'd also be good to note these errors alongside errors that OPAQUE generates on its own.

draft-irtf-cfrg-opaque.md
chris-wood: Maybe ServerAuthenticationMACError then?

draft-irtf-cfrg-opaque.md
bytemare: To me, it's not a valid point in the group. Would another name be more suitable?

bytemare: Note that it's different from the deserialization: an input can correctly deserialize, but if the config dictates a different group, than the deserialized input would not be valid for that config

chris-wood (approved):

chris-wood (approved):

PR #319, opened 2022-01-23
Match `generate_ke1` and `generate_ke3` API to document

jamesw1892:
With client transcript creation split between `generate_ke1` and `generate_ke3`, extra arguments like `idU`, and `pkU` are required for `generate_ke1`. However, if done entirely in `generate_ke3` then they are only required for `generate_ke3`.

This brings the function arguments in line with the pseudocode in the document. Also, in the document, transcript (preamble) creation is done entirely in `ClientFinalize`: https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-07.html#name-3dh-client-functions

Also, this way `pkS` is not required for `generate_ke1` or `generate_ke3` since it is obtained from `core.recover_credentials` in `generate_ke3`.

I can't think of a reason why the transcript creation needs to be split across the functions, but please let me know if there is. I hope you don't mind me submitting these pull requests! I've been trying to write a toy OPAQUE client extending this, but coudn't work out how to get `pkS` to give to `generate_ke1` without sending it separately to all the messages described in the document - after this change, I won't need it.

### DIFF
--- a/poc/opaque_ake.sage
+++ b/poc/opaque_ake.sage
@@ -108,7 +108,7 @@ class OPAQUE3DH(KeyExchange):
 
         return server_mac_key, client_mac_key, session_key, handshake_secret
 
-    def generate_ke1(self, pwdU, idU, pkU, idS, pkS):
+    def generate_ke1(self, pwdU):
         cred_request, cred_metadata = self.core.create_credential_request(pwdU)
         serialized_request = cred_request.serialize()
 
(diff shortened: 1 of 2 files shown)


bytemare:
Hi @jamesw1892,

The transcript hashing is split because (I think) it allows to implicitly keep the arguments in the running hash, instead of carrying the state outside the function. It's not actually hashing, just adding input to the buffer.

As you correctly identified, this behaviour is not described in the document, and rightfully so: the source of truth of the protocol specification is the document, and the client doesn't have access to some of these inputs.
The reference Python/Sage implementation is primarily there to provide proof-of-concept and generate test vectors, and so takes some liberties in its inner workings. On the other hand, our Go and Rust implementations hash the transcript in KE3 only.

Could you open an issue first, next time you have questions or suggestions? So we can discuss it there before diving into PRs :)

Nevertheless, I think your point is valid, the POC should reflect the spec more closely, be it only to avoid confusion for implementers.

jamesw1892:
Hi @bytemare 

Thank you for looking at it, understood. I will certainly submit an issue next time. Would you like me to convert this into an issue too?

bytemare:
No, it's ok :)

Let's wait to see if @chris-wood and @kevinlewi have an opinion on this

chris-wood:
I'm OK with PRs with out issues! And I think the intuition to treat the reference implementation as an extension of the spec is good. (We can probably go _further_ and make sure things are aligned.)

chris-wood (approved):

PR #322, opened 2022-01-25
Updating submodules in poc

kevinlewi:


bytemare (approved):

bytemare (changes requested):
If we merge this PR as is, it will break the vectors. We should adapt to the OPRF update first

bytemare:
I'm on it 👍

PR #327, opened 2022-02-02
Promote enumeration

### LINKED ISSUE #253: Promote discussion of client enumeration attacks
We don't discuss this until we get to CreateCredentialResponse, so it comes out of nowhere. Let's see if we can bubble this up to the intro or protocol overview.

chris-wood:
Closes #253.

cc @hugokraw for awareness

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -198,7 +198,7 @@ protocols"
     seriesinfo: http://eprint.iacr.org/2020/313
     date: 2020
 
-  SIGNAL:
+  3DH:
     title: "Simplifying OTR deniability"
     seriesinfo: https://signal.org/blog/simplifying-otr-deniability
     date: 2016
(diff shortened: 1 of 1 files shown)


hugokraw:
Two nits regarding terminology (just looking at this pull request changes):
- The term "offline registration" seems to indicate that it is different than "online registration" - in actuality, in most case registration is "online".
- "interacting with the server to enumerate clients registered via the first phase": I wonder how many people understand this use of the term "enumerate" - the attack is indeed called "enumeration attack" but the use of "enumerate" in seemingly non-technical English may not be obvious.

chris-wood:
> The term "offline registration" seems to indicate that it is different than "online registration" - in actuality, in most case registration is "online".

This was made to align terminology elsewhere in the draft. If we want to drop the "offline" and "online" distinction, we can do so in a followup PR.

chris-wood:
> "interacting with the server to enumerate clients registered via the first phase": I wonder how many people understand this use of the term "enumerate" - the attack is indeed called "enumeration attack" but the use of "enumerate" in seemingly non-technical English may not be obvious.

This is a fair point. I updated the text to explain what enumeration is before using that term.

bytemare (approved):

PR #329, opened 2022-02-02
Use the OPRF DeriveKeyPair function and update OPRF usage

chris-wood:
This change uses the DeriveKeyPair function from the OPRF spec for both the OPRF key generation and authentication key generation. It also adds serialization steps to the pseudocode and reference implementation, as that is no longer handled by the OPRF API.

Closes #328.
Closes #241.

PR #332, opened 2022-02-03
added libopaque to implementations

stef:
conforms to testvectors before the voprf09 update. proof:
https://github.com/stef/libopaque/blob/master/src/tests/opaque-testvectors.c

has one deviation, did not implement the voprf evaluate function in anticipation of it going back to the original oprf variant. hence the output for that is hardcoded, but will be removed with the vporf09 update.


chris-wood:
Awesome! Should we hold this while #329 is ironed out?

stef:
i adapted to the changes in #329 and conform: https://github.com/stef/libopaque/commit/9d4aed3e65cb987527ef7298f366bf30fdd158a2

chris-wood:
Lovely :) Merging this as well!

PR #333, opened 2022-02-05
Refactor error and implementation considerations

chris-wood:
This change refactors the error and implementation considerations as presented in #314, and also leaves a slot for us to address #312 in a followup change.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1671,7 +1671,7 @@ server-authenticated channel over which OPAQUE registration and login is run.
 ## Error Considerations
 
 Some functions included in this specification are fallible. For example, the
-authenticated key exchange protocol may fail because client's password was
+authenticated key exchange protocol may fail because the client's password was
 incorrect or the authentication check failed, yielding an error. The explicit
 errors generated throughout this specifiation, along with conditions that lead
 to each error, are as follows:
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -736,7 +736,7 @@ Output:
 - export_key, an additional client key.
 
 Exceptions:
-- InvalidEnvelopeMACError, the envelope fails to be recovered.
+- EnvelopeRecoveryError, the envelope fails to be recovered.
 
 def Recover(randomized_pwd, server_public_key, envelope,
             server_identity, client_identity):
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -1489,7 +1489,7 @@ Output:
 - session_key, the shared session secret.
 
 Exceptions:
-- ServerAuthenticationMACError , the handshake fails.
+- ServerAuthenticationError , the handshake fails.
```
bytemare, line 1492: I thought we agreed on your suggestion here
https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/314#discussion_r798821465

draft-irtf-cfrg-opaque.md
```diff
@@ -1639,87 +1639,67 @@ applications can use to control OPAQUE:
...
-| Name                        | Role   | Stage      | Reason                                         |
-|:----------------------------|:-------|:-----------|:-----------------------------------------------|
-| InvalidServerSecretKeyError | Server | ServerInit | Invalid server secret key.                     |
-| InvalidServerPublicKeyError | Server | ServerInit | Invalid server public key.                     |
-| InvalidOPRFSeedLengthError  | Server | ServerInit | Invalid OPRF seed length.                      |
-| InvalidEnvelopeLengthError  | Server | ServerInit | Invalid envelope length for the configuration. |
+## Error Considerations
+
+Some functions included in this specification are fallible. For example, the
+authenticated key exchange protocol may fail because client's password was
```
bytemare, line 1674:
```suggestion
authenticated key exchange protocol may fail because the client's password was
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1639,87 +1639,67 @@ applications can use to control OPAQUE:
...
+- ClientAuthenticationError: The server failed to complete the authenticated
+  key exchange protocol with the client; {{ake-server}}.
+
+Beyond these explicit errors, OPAQUE implementations can produce implicit errors.
+For example, if protocol messages sent between client and server do not match
+their expected size, an implementaton should produce an error. More generally,
+if any protocol message received from the peer is invalid, perhaps because the
+message contains an invalid public key (indicated by the AKE DeserializeElement
+function failing) or an invalid OPRF element (indicated by the OPRF DeserializeElement),
+then an implementation should produce an error.
```
bytemare, line 1692: I think we should also mention implicit errors originating from VOPRF (e.g. InvalidInputError) and hash-to-curve (if any)

draft-irtf-cfrg-opaque.md
chris-wood: Upon rereading things and making this PR I fell back to my original suggestion. I don't think it's important that the application know it's the MAC that failed here, and it's only guidance anyway. (An implementation could choose to be more specific.)

draft-irtf-cfrg-opaque.md
chris-wood: `DeserializeElement` is one of those implicit OPRF errors. I don't think we need to be exhaustive with the list of implicit things here.

I plan to address InvalidInputError and how OPAQUE should deal with it in response to #312.

PR #338, opened 2022-02-11
Correct `oprf_seed` length in POC test

jamesw1892:
An `oprf_seed` is generated 3 times in `test_opaque_ake.sage`. The other two times, the correct length, `Nh` (hash digest size), is used, but here, `Nok` (OPRF private key size) was used instead.

In `default_opaque_configuration`, these are different so it does make a difference.

The draft states it should be `Nh`, e.g.: in this paragraph: https://www.ietf.org/archive/id/draft-irtf-cfrg-opaque-07.html#section-5-1


stef:
i suppose this change also affects all test vectors for this test-case, so these need to be adjusted as well?

jamesw1892:
> i suppose this change also affects all test vectors for this test-case, so these need to be adjusted as well?

Yes it does, thank you for spotting!

chris-wood (approved):
Thanks! Can you please also copy these updated vectors (in `poc/vectors/formatted.txt`) into the main specification file?

jamesw1892:
> Thanks! Can you please also copy these updated vectors (in `poc/vectors/formatted.txt`) into the main specification file? If it's easier for you, I can do that in a followup change.

@chris-wood Done, I didn't realise they needed to be updated there too.

jamesw1892:
> No problem =) By the way, @jamesw1892, are you working on an implementation? If so, can I encourage you to add it to our [list of implementations](https://github.com/cfrg/draft-irtf-cfrg-opaque#implementations)?

@chris-wood I'm not sure if you would class it as an implementation. It's just an extension of this reference implementation to a client and server program to make it clear what needs to be done by each party. It doesn't re-write functionality in the reference implementation. You can see it here if interested: https://github.com/jamesw1892/OPAQUE/tree/master/poc#Usage

chris-wood:
@jamesw1892 @stef FYI when I generate test vectors based on this branch locally I get results that differ from what's been merged. @jamesw1892 is it possible your submodule references were behind? (I'll fix this in a new PR.)

jamesw1892:
@chris-wood I'm not sure what has caused it, but if you get different test vectors, definitely go with what you get. I thought my submodules were up to date but am fairly new to submodules. I generated the test vectors from my OPAQUE repo where I have made other changes, but I don't think I have made changes specifically to any code used to generate the test vectors.

PR #340, opened 2022-02-13
MHF to KSF throughout

chris-wood:
And fix an issue with the generated vectors in #338. And update to the latest voprf submodule while we're at it.

cc @stef, @jamesw1892


bytemare (approved):
Checked :) lgtm

stef:
i'll have a look later tonight.

stef:
i'm confused what exactly caused the change of the testvectors? as far as i see @jamesw1892 change f90a96e60d681c0f888d0bb93558b898376ff69d only affected: https://github.com/jamesw1892/OPAQUE/blob/ebebf8b106fcd008b3c820baa73dfca1e7dd8721/poc/test_opaque_ake.sage#L94 which should not affect the very first set of test vectors, since not even the pwdU (iloveopaque) matches in that function with the pwdU in the first testvector (batteryhorsestaple...) i looked at this PR but couldn't find why the test vectors changed. i also looked here, maybe the VOPRF changed in any way, but even here https://github.com/cfrg/draft-irtf-cfrg-voprf/compare/f1f8e0fb1f1d1eceb6cd40875240289b514c1674...015e34156e33daf1d75b693064952aed0f694f20 i couldn't find anything. would you please point out what changed, where and why? it would also be great in the future to mark commits that change the test vectors so that it is easier to find out why.

chris-wood:
I think his test vectors were run with an older version of the VOPRF repository. I ran them with the updated version.

stef:
Ok, but why did the testvectors change since the last time? i'm looking at this: https://github.com/cfrg/draft-irtf-cfrg-opaque/commits/eb699ddfb91f64ea90c9ed5463b0e7fa1e619acc/poc/vectors/vectors.json - and in this list my impl is conforming to eae5040eec78f69eaf77a3e7b59900e40f1899ef, after that there is only jamesw1892 change, and the change from this PR here. i when i run compare between these commits i get this: https://github.com/cfrg/draft-irtf-cfrg-opaque/compare/eae5040eec78f69eaf77a3e7b59900e40f1899ef..eb699ddfb91f64ea90c9ed5463b0e7fa1e619acc i am unable to see where there is a change that mandates the change of the text vectors. maybe it's lost in the noise, please help see me what i fail to see

chris-wood:
I'm not really sure how to clarify this further for you. The oprf_seed change did change the test vectors, but it appears as though the test vectors produced in https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/338 used an incorrect version of the voprf submodule since. So this PR fixed that.

stef:
yes the oprf_seed change did affect the testcase: "test_registration_and_authentication()" which uses "iloveopaque" as a pwdU, so that should leave the testvector where the password "batteryhorse..." alone and unaffected, and yet even there the testvectors changed.

stef:
ooooh i get it. the changed testcase is called before the testcase that i'm verifying, so that means that the random byte stream is offset... can we please just fix the random bytestream instead please?

chris-wood:
Yes. That will happen in a future change.

stef:
test vectors should never change without a good reason. please?

stef:
this causes a lot of extra and unnecessary work for implementers.

chris-wood:
That’s a fine ideal, but we’ve never actually made that commitment. As I said, we will do our best to see that this is done. Patience, please.

stef:
the cleanest would be of course if the poc would read the input values from the testvectors itself.

stef:
in the short-run i think the cheapest would be to just read Nok bytes from the rng to restore the stream, and use that value twice so it gets the expected NH length.

stef:
i mean in jamesw1892 one line fix.

PR #342, opened 2022-02-14
fix deviation of the test-vectors

stef:
see discussion in https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/340


chris-wood (changes requested):
Thanks for the addition, but this isn't the right fix to the problem. We need to change how randomness is plumbed into the protocol participants, and I'll do that soon!

stef:
i agree, it's not the right fix for the generic problem. but it is a quick and dirty fix for the last instantiation of it.

PR #343, opened 2022-02-15
Add opaque-ts TypeScript library.

armfazh:
A TypeScript implementation can be found at https://github.com/cloudflare/opaque-ts
which is complaint with published [draft-irtf-cfrg-opaque-07](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-opaque-07).


chris-wood (approved):

chris-wood:
Thanks! 🎉

PR #344, opened 2022-02-16
Plumb fixed randomness into each test vector.

chris-wood:
This change ensures that each test vector is run with the same fixed seed that is used to initialize a randomness seed that's distinct for each client and server. (Client and server share the same randomness during registration, but we can easily change that if desired.) We can also use a different seed for each test vector if desired.

cc @stef, @armfazh


armfazh:
I ran `make vectors` and obtained the same test vectors.

stef:
awesome! however there is still a problem if within a testcase the size/order/number of inputs changes. that will also throw of the rng output. i understand that it's difficult to hardcode the input vectors especially in the submodules of voprf and h2c. but then, that's exactly what we implementers also do in our code where we verify the testvectors.

stef:
i'm gonna check this tomorrow.

chris-wood:
> awesome! however there is still a problem if within a testcase the size/order/number of inputs changes. that will also throw of the rng output. i understand that it's difficult to hardcode the input vectors especially in the submodules of voprf and h2c. but then, that's exactly what we implementers also do in our code where we verify the testvectors.

Yeah, I'm open to suggestions to hear how we might improve this using the new plumbing. @stef, when you have a moment, can you point me to your tests that you wrote?

stef:
sure, here is my test code: https://github.com/stef/libopaque/blob/master/src/tests/opaque-testvectors.c - and in my opaque implementation https://github.com/stef/libopaque/blob/master/src/opaque.c i have all over the place ifdefed code like this:
```
#ifdef CFRG_TEST_VEC
  // "oprf_key": "3dc1f2c5212510244924ab97b10c19f8b0f0d9444295de5e7d2c9b9f8f8edf09",
  unsigned char rtest[] = {
       0x3d, 0xc1, 0xf2, 0xc5, 0x21, 0x25, 0x10, 0x24,
       0x49, 0x24, 0xab, 0x97, 0xb1, 0x0c, 0x19, 0xf8,
       0xb0, 0xf0, 0xd9, 0x44, 0x42, 0x95, 0xde, 0x5e,
       0x7d, 0x2c, 0x9b, 0x9f, 0x8f, 0x8e, 0xdf, 0x09
  };
  unsigned int rtest_len = 32;
  memcpy(kU,rtest,rtest_len);
#else
  crypto_core_ristretto255_scalar_random(kU);
#endif
```

you can imagine how much effort it is to always update the testvectors when them change...

chris-wood:
Thanks! Are you open to PRs against that repo to try and make this a bit easier and more automated?

stef:
sure, but this affects at least also aldens ecc implemenation, cannot imagine that the others do not suffer also from similar problems, but i haven't looked at the other impls...

chris-wood:
> sure, but this affects at least also aldens ecc implemenation, cannot imagine that the others do not suffer also from similar problems, but i haven't looked at the other impls...

Yep, for sure. I'm trying to figure out what's the best approach here that balances everyone's concerns. I'll hold this PR open until we get a chance to look at other implementations and how they're dealing with test vectors.

kevinlewi:
In the opaque-ke implementation, we have inner functions which are called by both the "real implementation" and the testing of the test vectors, for instance, when the real implementation has to sample randomness, but the test vectors already specify the randomness.

Then, the test cases just parse the test vectors and call these inner functions with the corresponding arguments. Changing the test vectors is not too hard to deal with because we just dump them into this file: https://github.com/novifinancial/opaque-ke/blob/main/src/tests/opaque_vectors.rs

and they get parsed and the tests are run here: https://github.com/novifinancial/opaque-ke/blob/main/src/tests/test_opaque_vectors.rs

So maybe each implementation needs to write a parser in order to keep up with the latest test vectors.

stef:
btw is there any reason that the inputs have to be totally random? can they not be some constant? or some constants? that would make things much easier on all accounts...

stef:
for example i have this for my own testcases:

```
#ifdef NORANDOM
void a_randombytes(void* const buf, const size_t len) {
  size_t i;
  for(i=0;i<len;i++) ((uint8_t*)buf)[i]=i&0xff;
}
````

stef:
i added a new tool to libopaque which generates a header file with the test vectors and thus makes updating them much easier: https://github.com/stef/libopaque/commit/74207f7459236c4b687bc53da4e1c1eca404d817

using this i was able to verify the testvectors in this PR.

having said this, i think that input test vectors should only change if there is a good reason for this, either because the specification document changes and mandates this, or if there was a bug in the poc generating the test vectors, but even then the number of vectors that change should be kept to a minimum. this is useful for us to detect any errors, which would be hidden if the input vectors change.

bytemare (approved):

bytemare:
My approach is to use the `vectors.json` file directly. In my tests, I parse the file, load the configuration, use the input values, run the protocol for the multiple suites, and during execution I compare with the intermediate and output values. The testing calls the external API and verifies internal intermediate values (I do the same for VOPRF).
Here's that code: https://github.com/bytemare/opaque/blob/main/tests/vectors_test.go

In this instance, I just had to replace the `vectors.json` file and it executed smoothly, I didn't have to change anything in my implementation or vectors.

PR #353, opened 2022-04-28
Auth ClientFinalize prototype should return export_key

cdrubin:
Update pseudocode prototype for function ClientFinalize to match the rest of the document which has it return the `export_key`


kevinlewi (approved):
LGTM, thanks for the catch!

PR #352, opened 2022-04-19
Removing AKE deps section and defining KE1/KE2/KE3 structs

kevinlewi:
As discussed offline, this will remove the "Authenticated Key Exchange (AKE) Protocol" subsection from the "Cryptographic Dependencies" section. This also means removing the definitions of the subroutines "AuthInit", "AuthResponse", "AuthFinish".

This is also an opportunity to clearly define the structs KE1, KE2, and KE3 in the "AKE Messages" section.

Partially addresses #336, but I think in order to complete #336, we should also re-order the sections in the Online AKE stage slightly, so that we aren't at any point referencing things that haven't yet been defined. This means moving the "Credential Retrieval" subsection up top, followed by "AKE messages", followed by "Client Authentication Functions" and "Server Authentication Functions". Not addressing this reordering in this PR, but will do so in a follow-up one.

PR #355, opened 2022-05-17
Reordering AKE subsections

kevinlewi:
This finishes the last portion of #352, in which I wanted to reorder the subsections slightly in the AKE portion to make things more clear.

Before this change, we had the following ordering of subsections in AKE:

"Online Authenticated Key Exchange" main section with protocol description at a high level 
- "Client Authentication Functions" subsection
- "Server Authentication Functions" subsection
- "Credential Retrieval" subsection
- "AKE Protocol"
  - "AKE Messages" section, this is where KE1, KE2, KE3 are defined
  - "3DH Client Functions" subsection
  - "3DH Server Functions" subsection

So the main problem was that the first two subsections, Client/Server Authentication Functions, were outputting messages that were only defined in the "AKE Messages" subsection, which was much later.

This seems like quite a different structure from the Registration portion, which is currently organized like:

"Client Credential Storage and Key Recovery" main section
"Offline Registration" main section
- "Registration Messages" subsection
- "Registration Functions" subsection

which is a way simpler layout.

--------

So the change I am making here is to:
1) Combine the "Client Authentication Functions" and "Server Authentication Functions" into one subsection, called "AKE Functions", so that it matches up with the layout of registration
2) Put "AKE Messages" as the section immediately following the main protocol description, followed by the new "AKE Functions" subsection
3) Leave the remainder of the subsections as-is in terms of ordering. So the new layout looks like this now:

"Online Authenticated Key Exchange" main section with protocol description at a high level 
- "AKE Messages" subsection
- "AKE Functions" subsection
- "Credential Retrieval" subsection
- "AKE Protocol"
  - "3DH Client Functions" subsection
  - "3DH Server Functions" subsection
  
  Hope this makes sense!
  
  In making this change, there were also some smaller adjustments I made, as well:
  - renamed the 3dh functions from "Start" to "AuthStart", "Response" to "AuthRespond" (it should be at least be a verb), "ClientFinalize" to "AuthClientFinalize", and "ServerFinalize" to "AuthServerFinalize"
  - I also added a definition for ServerFinish, which simply calls AuthServerFinalize, but this was previously a completely undefined function
  - Added text at the beginning of the "AKE Functions" subsection to clarify that the subroutines that we call there are defined in later sections

Btw, a change I was tempted to make, but did not end up making was to combine the "3DH client functions" and "3DH server functions" also into one section. We don't seem to be doing the splitting of client vs. server functions elsewhere anymore (especially after this change), so it is necessary that we do it in the very last section?

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -727,6 +727,7 @@ The registration protocol then runs as shown below:
               ------------------------->
 ~~~
 
+{{registration-messages}} describes the formats for the above messages, and
 {{registration-functions}} describes details of the functions and the
 corresponding parameters referenced above.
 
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -954,11 +955,89 @@ The server state may have the following fields:
...
+of fixed size `Nm`, computed using `Km2`, defined below.
+
+## AKE Functions {#ake-functions}
+
+In this section, we define the main functions used to produce the AKE messages
+in the protocol. Note that this section relies on definitions of subroutines defined
+in later sections:
+- `CreateCredentialRequest`, `CreateCredentialResponse`, `RecoverCredentials`
+  defined in {{cred-retrieval}}
+- `AuthStart`, `AuthRespond`, `AuthClientFinalize`, and `AuthServerFinalize`
```
chris-wood, line 1037: Should we make `AuthStart` be `AuthClientStart` (and similar for AuthRespond) so the roles are explicit?

chris-wood (approved):
LGTM with one suggestion for a rename. This is indeed a better layout in my opinion!

draft-irtf-cfrg-opaque.md
kevinlewi: Good idea! Updated in the latest commit.

PR #362, opened 2022-05-30
Separates ClientState and ServerState from ClientAkeState and ServerAkeState

### LINKED ISSUE #291: Determine the content of client/server states during online authentication
Currently the state is vaguely specified. Specification must be clearer in what is the content of the state. 
(this may need some minor fixes, but still needs to be addressed).

Raised after:

_Originally posted by @chris-wood in https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/286#discussion_r751669794_

kevinlewi:
Closes #291.

Adding in formal definitions for `ClientState`, `ServerState`, `ClientAkeState`, and `ServerAkeState`.

Note that for the password field, which can be variable-length, I set the type to `opaque password<1..2^16-1>`, even though this is not a strict requirement. I also set blind to be `opaque blind<1..2^16-1>`. Let me know if these should be changed!

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -956,25 +956,12 @@ material for the OPRF and AKE, `client_state` and `server_state`, respectively.
 
 The client state `ClientState` may have the following fields:
 
-```
-struct {
-  opaque password<1..2^16-1>,
-  opaque blind<1..2^16-1>,
-  ClientAkeState client_ake_state,
-} ClientState;
-```
 
 - password: The client's password.
 - blind: The random blinding inverter returned by `Blind()`.
 - client_ake_state: The `ClientAkeState` defined in {{ake-protocol}}.
 
 The server state `ServerState` may have the following fields:
-
-```
-struct {
-  ServerAkeState server_ake_state,
-} ServerState;
-```
 
 - server_ake_state: The `ServerAkeState` defined in {{ake-protocol}}.
 
@@ -1362,29 +1349,14 @@ This section describes the authenticated key exchange protocol for OPAQUE using
 3DH, a 3-message AKE which satisfies the forward secrecy and KCI properties
 discussed in {{security-considerations}}.
 
-The client AKEstate `ClientAkeState` mentioned in {{online-phase}} has the
+The client AKE state `ClientAkeState` mentioned in {{online-phase}} has the
 following fields:
-
-```
-struct {
-  opaque client_secret[Nsk],
-  KE1 ke1,
-} ClientAkeState;
-```
 
 - client_secret: An opaque byte string of length `Nsk`.
 - ke1: A value of type `KE1`.
 
 The server AKE state `ServerAkeState` mentioned in {{online-phase}} has the
 following fields:
-
-```
-struct {
-  opaque expected_client_mac[Nm],
-  opaque session_key[Nx],
-} ServerAkeState;
-```
-
 - expected_client_mac: An opaque byte string of length `Nm`.
 - session_key: An opaque byte string of length `Nx`.
 
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -954,15 +954,29 @@ The protocol runs as shown below:
 Both client and server may use implicit internal state objects to keep necessary
 material for the OPRF and AKE, `client_state` and `server_state`, respectively.
 
-The client state may have the following fields:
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -954,15 +954,29 @@ The protocol runs as shown below:
...
-The client state may have the following fields:
+The client state `ClientState` may have the following fields:
+
+```
+struct {
+  opaque password<1..2^16-1>,
+  opaque blind<1..2^16-1>,
+  ClientAkeState client_ake_state,
+} ClientState;
+```
```
chris-wood, line 965:
```suggestion
```

draft-irtf-cfrg-opaque.md
```diff
@@ -954,15 +954,29 @@ The protocol runs as shown below:
...
+- client_ake_state: The `ClientAkeState` defined in {{ake-protocol}}.
 
-The server state may have the following fields:
+The server state `ServerState` may have the following fields:
+
+```
+struct {
+  ServerAkeState server_ake_state,
+} ServerState;
+```
```
chris-wood, line 977:
```suggestion
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1344,15 +1362,29 @@ This section describes the authenticated key exchange protocol for OPAQUE using
 3DH, a 3-message AKE which satisfies the forward secrecy and KCI properties
 discussed in {{security-considerations}}.
 
-The AKE client state `client_ake_state` mentioned in {{online-phase}} has the
+The client AKEstate `ClientAkeState` mentioned in {{online-phase}} has the
```
chris-wood, line 1365:
```suggestion
The client AKE state `ClientAkeState` mentioned in {{online-phase}} has the
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1344,15 +1362,29 @@ This section describes the authenticated key exchange protocol for OPAQUE using
...
 following fields:
 
+```
+struct {
+  opaque client_secret[Nsk],
+  KE1 ke1,
+} ClientAkeState;
+```
+
```
chris-wood, line 1374:
```suggestion
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1344,15 +1362,29 @@ This section describes the authenticated key exchange protocol for OPAQUE using
...
+The server AKE state `ServerAkeState` mentioned in {{online-phase}} has the
 following fields:
 
+```
+struct {
+  opaque expected_client_mac[Nm],
+  opaque session_key[Nx],
+} ServerAkeState;
+```
+
```
chris-wood, line 1387:
```suggestion
```

chris-wood (changes requested):
The actual data structure is an implementation-specific decision and should not be mandated like this, so I think we ought to remove some details and just say list the things that can be referenced.

stef:
i agree with Chris that these are implementation specific. so **none** of this should be binding. however i see value
  - for implementers as a kind of checklist,
  - for interoperability for suggesting minimum/maximum size for variable sized fields - this would enable converters between different implementations,
  - most importantly lifetime, sensitivity/confidentiality of these fields (for example can they be serialized to storage, do they need to be encrypted, will they lose importance after the next step, etc)

i understand that much of this is already specified through-out of this document, and thus i think this is merely a convenience feature for implementers who don't have to piece this information together from various parts of the spec. i guess that's a lot of words for saying put this whole thing into a non-binding appendix.

kevinlewi:
Thanks for the feedback. I agree that it shouldn't be required for implementations to implement this state. However, shouldn't we still have struct definitions for these, since they are referenced in the functions? Just going off of the discussion in Issue #291 which brought this to our attention in the first place...

Comment from @armfazh :
> So, either formally specifying what the state contains, or removing at all the state structure are both good alternatives. Currently, it is specified something in between.

chris-wood:
> However, shouldn't we still have struct definitions for these, since they are referenced in the functions? 

No, I don't think so. It is sufficient to simply say what the data structure must contain. It is not necessary to say anything about how that data structure is implemented.

kevinlewi:
I see, got it. Alright, I committed your suggestions which removed the formal specifications. This PR then doesn't do too much, other than separate ClientState/ServerState from ClientAkeState/ServerAkeState. 

Are we still good to close #291 after this lands?

bytemare (approved):

PR #374, opened 2022-09-12
Rename "ephemeral key share" to "ephemeral public key"

### LINKED ISSUE #371: why keyshare why not ephemeral key pairs?
I'm confused by the term keyshare, this term doesn't tell me anything, and i think it obscures the nature of this value, the keyshare is nothing but an ephemeral key pair, and if we would call it that i think understanding would be easier.

maybe you use keyshare because it's a more generic term that applies to other opaque instantiations, but the current spec focuses on 3dh where the keyshare is really an ephemeral  keypair, please lets call it that.

kevinlewi:
Closes #371.

See the discussion in #371 for more context. Note that this only changes the prose describing the variable names, and not the variable names themselves.

cc'ing @stef for feedback

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1673,8 +1673,8 @@ parameters that are needed to prevent cross-protocol or downgrade attacks.
 
 Absent an application-specific profile, the following configurations are RECOMMENDED:
 
-- OPRF(ristretto255, SHA-512), HKDF-SHA-512, HMAC-SHA-512, SHA-512, Scrypt(32768,8,1), internal, ristretto255
-- OPRF(P-256, SHA-256), HKDF-SHA-256, HMAC-SHA-256, SHA-256, Scrypt(32768,8,1), internal, P-256
+- OPRF(ristretto255, SHA-512), HKDF-SHA-512, HMAC-SHA-512, SHA-512, Scrypt(32768,8,1), ristretto255
+- OPRF(P-256, SHA-256), HKDF-SHA-256, HMAC-SHA-256, SHA-256, Scrypt(32768,8,1), P-256
 
 Future configurations may specify different combinations of dependent algorithms,
 with the following considerations:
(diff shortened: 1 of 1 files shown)


bytemare (approved):

kevinlewi:
@stef: Personally, I'm reluctant to do the variable renaming because I think it is too big of a change at this point. But perhaps if others chime in with an opinion on whether or not it's worth the variable renaming, we could consider it? @bytemare @chris-wood

bytemare:
I'm not against, for consistency I think it's better to also rename the vectors.
(New implementors or readers might get a bit confused)

kevinlewi:
Here are the four variables we would need to rename:
- `server_keyshare`
- `client_keyshare`
- `server_private_keyshare`
- `client_private_keyshare`

For context, these names were originally decided on in https://github.com/cfrg/draft-irtf-cfrg-opaque/commit/87c3287dc789eda67dcb93fbd481334025c71cc6 , where their previous names were:
- `epkS`, `epkU`, `eskS`, `eskU`

Any proposals for what each of these should be renamed to? Some options that come to mind:
1) `client_e_pk`, `client_e_sk`, `server_e_pk`, `server_e_sk`
2) `client_ephemeral_public_key`, `client_ephemeral_private_key`, `server_ephemeral_public_key`, `server_ephemeral_private_key`

stef:
> I think it is too big of a change at this point. 

why do you think so? for existing implementations nothing changes. it helps clarity.

both of your proposals are good, but since this is a spec, clarity and expressiveness should trump brevity so i would prefer the 2nd. but i also have to admit that in my code i still use the names from 87c3287 - which i think despite their cryptic shortness are still more expressive than keyshare.

stef:
oh i see the sage ref impl needs to be changed i guess?

stef:
> IMO although the name keyshare may not be 100% illustrative as a variable name, it doesn't seem to significantly detract from a reader understanding the protocol (given that we also define the variable as being the ephemeral public/private key already).

true, it only confuses when working with the test vectors.

chris-wood (approved):

PR #377, opened 2022-10-25
Fix inaccuracies in function definitions

n-l-i:


kevinlewi (approved):
This is great. Thank you very much for catching these and making the contribution!

PR #379, opened 2022-11-28
Add missing close paren to Recover()

soatok:


bytemare:
Nice catch! Thank you @soatok !

I don't think this needs further review, merging.

PR #383, opened 2023-02-06
Adapt OPRF calls to take into account new IDs

bytemare:
This updates the dependencies to the new VOPRF updates and therefore updates the vectors.


kevinlewi (approved):
Awesome!

chris-wood (approved):

PR #385, opened 2023-02-24
Use the correct request in CreateCredentialResponse

fmarier:
There are two `request` members inside the `KE1` struct. The one that `CreateCredentialResponse()` needs is `credential_request`.


bytemare (approved):

bytemare:
Good catch!

PR #396, opened 2023-03-04
Fix h2c submodule

armfazh:
Changes:
 - Removes duplicated submodule for h2c. Better to use the one inside the voprf module.
 - ~Exports more values to make the Fake case reproducible. No changes on test vectors, just more verbose info for the `Fake=True` cases.~

### DIFF
--- a/poc/opaque_ake.sage
+++ b/poc/opaque_ake.sage
@@ -61,7 +61,7 @@ class OPAQUE3DH(KeyExchange):
         return {
             "Name": "3DH",
             "Group": self.config.group.name,
-            "OPRF": to_hex(I2OSP(self.config.oprf_suite.identifier, 2)),
+            "OPRF": self.config.oprf_suite.identifier,
             "KDF": self.config.kdf.name,
             "MAC": self.config.mac.name,
             "Hash": self.config.hash().name.upper(),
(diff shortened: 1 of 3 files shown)


bytemare (approved):
Works for me ! Thank you :)

kevinlewi (changes requested):
@armfazh: Thanks! I'm struggling to understand the need for the changes to make the fake case reproducible, though...

Why do we need to include the input parameters to generate KE1 (e.g. password, client_nonce, etc.), and view KE1 as an output (as opposed to an input)? Shouldn't KE1 be an input, and KE2 be the only output?

armfazh:
It wasn't required at all. Test vectors are ok as is.
I just update this PR with changes only for h2c submodule.

kevinlewi (approved):
LGTM! Thanks!

poc/opaque_ake.sage
```diff
@@ -61,7 +61,7 @@ class OPAQUE3DH(KeyExchange):
         return {
             "Name": "3DH",
             "Group": self.config.group.name,
-            "OPRF": to_hex(I2OSP(self.config.oprf_suite.identifier, 2)),
+            "OPRF": self.config.oprf_suite.identifier,
```
bytemare, line 64: Good catch! I must have forgotten to commit that

chris-wood (approved):

chris-wood:
Thanks, @armfazh! Merging.

Issue #390, opened 2023-02-26
Use of server_private_key in offline registration


mitar:
In [5. Offline Registration](https://cfrg.github.io/draft-irtf-cfrg-opaque/draft-irtf-cfrg-opaque.html#name-offline-registration-2) it lists `server_private_key` as input, but it is not used anywhere in the registration protocol shown later on.

bytemare:
Indeed! Good catch !

PR #397, opened 2023-03-08
Clarifications about Threshold OPRF

### LINKED ISSUE #389: Distributing the server
In [10.11 Password Salt and Storage Implications](https://cfrg.github.io/draft-irtf-cfrg-opaque/draft-irtf-cfrg-opaque.html#name-password-salt-and-storage-i) it says:

> OPAQUE enables defense against such offline dictionary attacks by distributing the server so that an offline attack is only possible if all - or a minimal number of - servers are compromised [[JKX18](https://cfrg.github.io/draft-irtf-cfrg-opaque/draft-irtf-cfrg-opaque.html#JKX18)].

My understanding is that the OPAQUE in the paper describes how one could distribute the server, but this spec itself does not mention this anywhere? (Or have I missed it.) Probably using a threshold OPRF? I think it would be beneficial to the spec if this would be more fleshed out somewhere. Or made into one of showcased features of OPAQUE. I think this is a cool feature but currently one finds hidden inside one set of parenthesis.

bytemare:
Addresses and closes #389 and #390


draft-irtf-cfrg-opaque.md
```diff
@@ -2127,10 +2126,11 @@ disclose their passwords to the server, even during registration. Note that a co
 server can run an exhaustive offline dictionary attack to validate guesses for the client's
 password; this is inevitable in any aPAKE protocol. (OPAQUE enables defense against such
 offline dictionary attacks by distributing the server so that an offline attack is only
-possible if all - or a minimal number of - servers are compromised {{JKX18}}.) Furthermore,
-if the server does not sample this OPRF key with sufficiently high entropy, or if it is not
-kept hidden from an adversary, then any derivatives from the client's password may also be
-susceptible to an offline dictionary attack to recover the original password.
+possible if all - or a minimal number of - servers are compromised {{JKX18}}. The details
+of this mechanism are out of scope of this document.) Furthermore, if the server does not
```
kevinlewi, line 2130: `out of scope of this document` -> `out of scope for this document`

draft-irtf-cfrg-opaque.md
```diff
@@ -2127,10 +2126,11 @@ disclose their passwords to the server, even during registration. Note that a co
 server can run an exhaustive offline dictionary attack to validate guesses for the client's
 password; this is inevitable in any aPAKE protocol. (OPAQUE enables defense against such
 offline dictionary attacks by distributing the server so that an offline attack is only
-possible if all - or a minimal number of - servers are compromised {{JKX18}}.) Furthermore,
-if the server does not sample this OPRF key with sufficiently high entropy, or if it is not
-kept hidden from an adversary, then any derivatives from the client's password may also be
-susceptible to an offline dictionary attack to recover the original password.
+possible if all - or a minimal number of - servers are compromised {{JKX18}}. The details
```
kevinlewi, line 2129: Can we keep it still as just one sentence?

`...servers are compromised {{JKX18}}. The details...`

-> 

`...servers are compromised {{JKX18}}, but the details...`

bytemare: Sure

draft-irtf-cfrg-opaque.md
bytemare: 👍

draft-irtf-cfrg-opaque.md
```diff
@@ -2127,10 +2126,11 @@ disclose their passwords to the server, even during registration. Note that a co
 server can run an exhaustive offline dictionary attack to validate guesses for the client's
 password; this is inevitable in any aPAKE protocol. (OPAQUE enables defense against such
```
chris-wood, line 2127: Since threshold OPAQUE is out of scope and, as far as I know, would require protocol changes, I would just remove this entire sentence ("(OPAQUE enables ... compromised {{JKX18}}.)") from the document.

chris-wood (changes requested):

draft-irtf-cfrg-opaque.md
stef: that depends on how you define protocol changes, yes a parameter to oprf_finalize would be larger (since it contains the blinded/evaluated shares that need to be combined before unblinding) but besides that the only thing that needs change is oprf_evaluate which acts as a proxy to the shareholders. but yeah that still doesn't put threshold in scope of this spec. which is a pity though (same for the oprf spec which lacks threshold variant, and would make this whole thing easier also in opaque). 

anyway i would leave mentioning this mitigation in here but declare it out of scope for this spec.

chris-wood: Minimally, clients would need to be aware of the different threshold servers and then generate messages for each. Right now, there is exactly one server in the protocol, and doing anything different seems like it would pretty clearly require a change.

stef: oh, actually! no, it is possible to do everything in oprf_evaluate! and nothing in the oprf_evaluate fn params would change, oprf_evaluate would simply act as a proxy that distributes the evaluation among the shareholders and then combines the result and returns that as the result of oprf_evaluate. from the opaque perspective this would be completely "opaque" and noone would have any idea if the oprf output is a threshold or non-threshold version. yes, of course oprf_evaluate would need to know about the shareholders and the threshold. but that is an oprf_evaluate internal setting and can be totally ignored by opaque.

stef: i will implement this asap!

chris-wood: That sort of deployment is indeed possible, but I don't think it faithfully captures the intent of the paper here. The proxy server can still be compromised, such as a single server could, to make everything fall apart. I claim that "true" benefits to adding threshold support only come if the client is actually a participant in that protocol.

stef: and i'm sure i will still satisfy the testvectors and will not have to change anything except for oprf_evaluate!

chris-wood: @stef see above -- I suspect we have different mental models for how one might use a threshold OPRF here.

stef: > That sort of deployment is indeed possible, but I don't think it faithfully captures the intent of the paper here. The proxy server can still be compromised, such as a single server could, to make everything fall apart. I claim that "true" benefits to adding threshold support only come if the client is actually a participant in that protocol.

hmmm right, if the opaque server and the threshold-oprf shareholder servers are all under the control of the attacker then this:

> server can run an exhaustive offline dictionary attack to validate guesses for the client's
password

will still work. but if the oprf shareholders are not under the attackers control, then it is a mitigation.

but then the other solution where the combination of the shares happens in oprf_finalize also makes no difference even though it happens at the client, since that can be also faked by the offline bruteforcing server.

stef: ok, ok, i'm holding my horses. but now i wanna know how to do this to claim the true benefits of a toprf. @hugokraw can you enlighten us please?

stef: hmmm, actually reading the paper:

> Threshold implementation. We comment on a simple extension of OPAQUE
> that can be very valuable in large deployments, namely, the ability to implement
> the OPRF phase as a Threshold OPRF [28]. In this case, an attacker needs to
> break into a threshold of servers to be able to impersonate the servers to the user
> or to run an offline dictionary attack. Such an implementation requires no user-
> side changes, i.e., the user does not need to know if the system is implemented
> with one or multiple servers.

this refers to:
S. Jarecki, A. Kiayias, H. Krawczyk, and J. Xu. TOPPSS: Cost-minimal password-protected secret sharing  based on threshold OPRF. In Applied Cryptology and Network Security – ACNS 2017, pages 39–58. Springer, 2017.

which says this:

> This contribution is based on the observation that a more efficient PPSS can result from replacing the  OPRF used in the protocols of [18,19] with its threshold (or multi-party) counterpart which we define as Threshold OPRF (T-OPRF).

stef: i just figured if we don't trust the opaque server then a dkg is necessary.

bytemare: @chris-wood I understand your point, but I advocate for still mentioning it since the paper talks about it.

The current phrase can be misleading, so I suggest adding to `10.1. Notable Design Differences` that this spec does not consider thresholding. What do you think?

chris-wood: I'd be OK mentioning it in the notable design differences section.

chris-wood: @bytemare will you be able to update this text?

bytemare: Please let me know what you think :)

draft-irtf-cfrg-opaque.md
```diff
@@ -1878,6 +1895,14 @@ implementation considerations.
   of the derivation of OPRF keys via a single PRF. As long as the derivation
   of different OPRF keys from a single OPRF has different PRF inputs, the
   protocol is secure. The choice of such inputs is up to the application.
+- {{JKX18}} comments on a strong defense against essentially the only
+  inherent remaining weakness of a "strong aPAKE", namely, offline
+  dictionary attacks upon server compromise. The authors suggest implementing
+  the OPRF phase as a Threshold OPRF {{TOPPSS}}, effecitvely forcing an
```
chris-wood, line 1901:
```suggestion
  the OPRF phase as a Threshold OPRF {{TOPPSS}}, effectively forcing an
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1878,6 +1895,14 @@ implementation considerations.
...
+  attacker to act online or to control at least t key shares, the threshold
+  number of keyshares necessary to recombine the shared key. This implementation
```
chris-wood, line 1903:
```suggestion
  attacker to act online or to control at least t key shares, where t is the threshold
  number of shares necessary to recombine the secret OPRF key. This implementation
```

chris-wood (approved):

chris-wood:
@bytemare can you run `make fix-lint` to tidy up whitespace?

draft-irtf-cfrg-opaque.md
```diff
@@ -1877,6 +1895,14 @@ implementation considerations.
   of the derivation of OPRF keys via a single PRF. As long as the derivation
   of different OPRF keys from a single OPRF has different PRF inputs, the
   protocol is secure. The choice of such inputs is up to the application.
+- {{JKX18}} comments on a strong defense against essentially the only
+  inherent remaining weakness of a "strong aPAKE", namely, offline
+  dictionary attacks upon server compromise. The authors suggest implementing
```
stef, line 1900: not only on server compromise, but also the server itself it is more than honest-but-curious.

bytemare: Thanks!

draft-irtf-cfrg-opaque.md
```diff
@@ -1878,6 +1895,14 @@ implementation considerations.
   of the derivation of OPRF keys via a single PRF. As long as the derivation
   of different OPRF keys from a single OPRF has different PRF inputs, the
   protocol is secure. The choice of such inputs is up to the application.
+- {{JKX18}} comments on a strong defense against essentially the only
+  inherent remaining weakness of a "strong aPAKE", namely, offline
+  dictionary attacks upon server compromise or honest-but-curious servers.
```
stef, line 1900: excuses! i was confusing when i wrote 
> more than honest-but-curious
which i intended to mean that the server is more "evil" than honest-but-curious. i guess, a better phrase would've been to just say: maliciuos. not sure if this really makes any difference, but in my understanding there is two cases: 1/ the server is compromised by an outside attacker who has then access to all the "records" and starts bruteforcing those and 2/ the server is operated by a malicious actor who is bruteforcing records entrusted to it. semantics,  schmemantics, dunno if it is necessary to complicate things in this regard. either way the server is compromised i guess. so let's leave it at that?
apologies again for being confusing and causing more work than needed.

draft-irtf-cfrg-opaque.md
```diff
@@ -1878,6 +1895,14 @@ implementation considerations.
...
+  The authors suggest implementing the OPRF phase as a Threshold OPRF {{TOPPSS}},
+  effectively forcing an attacker to act online or to control at least t key
+  shares, where t is the threshold number of shares necessary to recombine
+  the secret OPRF key. This implementation only affects the server and changes
```
hugokraw, line 1904: You should say explicitly that reconstructing the key does not represent a break of the system but only the ability to run offline dictionary attacks. That is, to be able to impersonate a user, an attacker needs to break into at least t OPRF servers and then run a dictionary attack against the specific user's password. Even breaking into all servers, still requires an exhaustive offline dictionary attack against the user's password. 
(Note to selves: It is a VERY STRONG protection.)

chris-wood, line 1904:
```suggestion
  the secret OPRF key and run an offline dictionary attack. This implementation only affects the server and changes
```

hugokraw:
Sorry, one more comment. I suggest the following edit  (marked in italics)

The authors suggest implementing the OPRF phase as a Threshold OPRF
{{TOPPSS}}, effectively forcing an attacker to act online or to control at
least t key shares, where t is the threshold number of shares necessary to
recombine the secret OPRF key*, and only then be able to* run an offline
dictionary attack.

You may or may not comment also in the following point:
If the OPRF servers are separate from the authentication server then
finding all n shares still does not help since you cannot run the dictionary
attack without the server's database.

draft-irtf-cfrg-opaque.md
bytemare: No problem :) I totally understand your point. If you find a way to express this clearly please don't hesitate to open a PR!

PR #401, opened 2023-03-20
Remove KSF parameters and mention config updates

### LINKED ISSUE #392: Rotating server keypair
I am missing mention/protocol/description of how can the server keypair be rotated. To me this is a bit surprising because I am used for keys to have expiration and are rotated before that expiration. OPAQUE seems to expect that server keys will be in use forever or at least be changed only after a compromise?

Currently, the only way I can see that rotation is done is by having support for that in clients so that they re-register after successful login. But clients might misbehave or simply skip that step. It looks to me like a potential deployment nightmare where some clients keep using old keypair.

### LINKED ISSUE #394: Migration to a new protocol version, hash, etc.
With regular password based authentication when the client sends the password to the server over TLS connection, migration to new ways of storing passwords is relatively easy: on successful login you migrate the user to new version or new hash, etc. Completely transparent to the user/client.

I am trying to understand how can this be done with OPAQUE. For example, how to bump hash complexity or pick a new hash? It looks like client would have to do that in some way (similar to #392).

Should a description of steps here be documented in the spec? Some recommended practice?

Maybe the solution is to always do both registration and login, which would probably address this, #392, and #388. In a way this is what sending plaintext password is: you can register the user, login the user, or even both at the same time, with the same type of a message (username and password) to the server.

### LINKED ISSUE #398: `params` in `Stretch`
When investigating what the right salt to use for Argon2 is, I noticed that the spec talks says the following:
> Stretch(msg, params): Apply a key stretching function with parameters `params` to stretch the input `msg` and harden it against offline dictionary attacks.

I'm not sure what `params` is supposed to be here and it's not specified anywhere else. I believe this should be removed from the spec?

---

For context:
I noticed that some OPAQUE implementations use a zero salt for Argon2, which I believe is incorrect. For example in Rust, Argon2 implementations sometimes **require** a salt, which according to the Argon2 specification is not correct. So OPAQUE implementations just used a zero salt.

### LINKED ISSUE #399: Include test vectors for Argon2id
Current test vectors use the Identity function as KSF.
while PoC implementation uses scrypt.

### LINKED ISSUE #400: what is the salt value for Argon2id?
In the parametrization for KSF=Argon2id there is no specification for the salt parameter expected by Argon2id.

According to [spec](https://www.password-hashing.net/argon2-specs.pdf): "...salt must be at least 8 bytes, and 16 for password hashing..."

chris-wood:
Closes #400
Closes #399 
Closes #398 
Closes #394 
Closes #392

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -320,6 +320,7 @@ The following functions are used throughout this document:
 - concat(x0, ..., xN): Concatenate byte strings. For example,
   `concat(0x01, 0x0203, 0x040506) = 0x010203040506`.
 - random(n): Generate a cryptographically secure pseudorandom byte string of length `n` bytes.
+- zeroes(n): Generate a string of `n` bytes all equal to 0 (zero).
 - xor(a,b): Apply XOR to byte strings. For example, `xor(0xF0F0, 0x1234) = 0xE2C4`.
   It is an error to call this function with arguments of unequal length.
 - ct_equal(a, b): Return `true` if `a` is equal to `b`, and false otherwise.
@@ -1679,9 +1680,9 @@ parameters that are needed to prevent cross-protocol or downgrade attacks.
 Absent an application-specific profile, the following configurations are RECOMMENDED:
 
 - ristretto255-SHA512, HKDF-SHA-512, HMAC-SHA-512, SHA-512,
-    Argon2id(t=1, p=4, m=2^21), ristretto255
+    Argon2id(S = zeroes(16), p=4, T = 32, m=2^21, t=1, v = 0x13, K = nil, X = "RFCXXXX", y = 2), ristretto255
 - P256-SHA256, HKDF-SHA-256, HMAC-SHA-256, SHA-256,
-    Argon2id(t=1, p=4, m=2^21), P-256
+    Argon2id(S = zeroes(16), p=4, T = 32, m=2^21, t=1, v = 0x13, K = nil, X = "RFCXXXX", y = 2), P-256
 
 Future configurations may specify different combinations of dependent algorithms,
 with the following considerations:
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1680,9 +1680,9 @@ parameters that are needed to prevent cross-protocol or downgrade attacks.
 Absent an application-specific profile, the following configurations are RECOMMENDED:
 
 - ristretto255-SHA512, HKDF-SHA-512, HMAC-SHA-512, SHA-512,
-    Argon2id(S = zeroes(16), p = 4, T = 32, m = 2^21, t = 1, v = 0x13, K = nil, X = "RFCXXXX", y = 2), ristretto255
+    Argon2id(S = zeroes(16), p = 4, T = Nh, m = 2^21, t = 1, v = 0x13, K = nil, X = nil, y = 2), ristretto255
 - P256-SHA256, HKDF-SHA-256, HMAC-SHA-256, SHA-256,
-    Argon2id(S = zeroes(16), p = 4, T = 32, m = 2^21, t = 1, v = 0x13, K = nil, X = "RFCXXXX", y = 2), P-256
+    Argon2id(S = zeroes(16), p = 4, T = Nh, m = 2^21, t = 1, v = 0x13, K = nil, X = nil, y = 2), P-256
 
 Future configurations may specify different combinations of dependent algorithms,
 with the following considerations:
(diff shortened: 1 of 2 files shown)


bytemare (approved):

daxpedda:
This doesn't address the issue in #400 and only partially addresses #398 (by removing `params`):
the question of what size of salt to use. Specifically to either use none if allowed by the spec or to use all zeroes.

chris-wood:
The point of this change is that the choice of KSF parameters is _entirely up the application_. It is not something that OPAQUE specifies. Thus, I believe it addresses all the issues referenced.

daxpedda:
I agree with that approach, imo OPAQUE should not deal with KSF parameters.

But there is a problem we might want to address one way or another: currently there is no consensus on the salt issue and it will eventually lead to incompatible implementations (e.g. https://github.com/facebook/opaque-ke/pull/275).

It would help to add a note just saying what to do here, which is actually still unclear. If I had nothing else but the spec to go by, I would not use a salt at all, which no implementation I know of does.

If it's still decided that we don't want to mention a salt at all in the spec, I would suggest an alternative solution. Currently Argon2 is the recommended algorithm, like the other recommended algorithms (P-*, Ristretto255), maybe adding test vectors would help. Which was requested in https://github.com/cfrg/draft-irtf-cfrg-opaque/issues/399, I assume the idea was rejected because we don't want to make it look like specific KSF parameters are suggested?

daxpedda:
In `opaque-ke` the output size is determined by the input size, which is the output size of the hash function used in the OPRF. So `Nh`, according to the spec.

I was wondering how that happened, but I found other implementations which did the same: [stef's implementation](https://github.com/stef/libopaque/blob/3ff26cd7fd5fa618361c33cf0c0c7e563a252a71/src/opaque.c#L208-L212), for example, did this too.

This is another point implementations have silently agreed on without the specification. This problem seems to get bigger as we go :laughing:.

I have to say that personally I don't have a stake in this, my use-cases don't require compatibility between multiple implementations, I'm using the same on both sides. Maybe somebody else with an actual stake in this will have more to say.

draft-irtf-cfrg-opaque.md
```diff
@@ -1677,8 +1680,9 @@ parameters that are needed to prevent cross-protocol or downgrade attacks.
 Absent an application-specific profile, the following configurations are RECOMMENDED:
 
 - ristretto255-SHA512, HKDF-SHA-512, HMAC-SHA-512, SHA-512,
-    Argon2id(t=1, p=4, m=2^21), ristretto255
-- P256-SHA256, HKDF-SHA-256, HMAC-SHA-256, SHA-256, Argon2id(t=1, p=4, m=2^21), P-256
+    Argon2id(S = zeroes(16), p = 4, T = 32, m = 2^21, t = 1, v = 0x13, K = nil, X = "RFCXXXX", y = 2), ristretto255
+- P256-SHA256, HKDF-SHA-256, HMAC-SHA-256, SHA-256,
+    Argon2id(S = zeroes(16), p = 4, T = 32, m = 2^21, t = 1, v = 0x13, K = nil, X = "RFCXXXX", y = 2), P-256
```
daxpedda, line 1685: That would add another point: associated data.

chris-wood: X is the associated data here.

daxpedda: Yes, what I meant to say is that this would another parameter that is not specified by the spec which will have to be silently agreed on to make implementations compatible.

Basically, until now associated data was left empty (I didn't check more implementations, but kevinlewi's and stef's is empty), now, implementations will probably use `RFCXXXX` as their associated data for all KSF's, even though the spec doesn't say you should do that.

chris-wood:
To hopefully avoid further nit picks, I just made the output length be `Nh` and set associated data to `nil`.

draft-irtf-cfrg-opaque.md
```diff
@@ -2275,7 +2292,9 @@ outputs computed during the authentication of an unknown or unregistered user. N
 All values are encoded in hexadecimal strings. The configuration information
 includes the (OPRF, Hash, KSF, KDF, MAC, Group, Context) tuple, where the Group
 matches that which is used in the OPRF. These test vectors were generated using
-draft-10 of {{OPRF}}.
+draft-21 of {{OPRF}}. The KSF used for each test vector is the identity function
```
armfazh, line 2295: May I suggest to use argon2id but with short `m`, for example instead `m = 2^21` set `m=8`.

kevinlewi (approved):
Sorry for the delays in getting to this review, but overall looks good to me!

Although I am probably just reiterating @chris-wood's points: I think that aligning implementations on the usage of OPAQUE with specific key stretching functions is outside of the scope of this spec.

This applies to mentioning anything about KSF salt length (for example), which some applications that use OPAQUE may want to align on... but I see as still being outside of the spec scope.

PR #402, opened 2023-03-21
Registration requires integrity

### LINKED ISSUE #393: Registration requires integrity protected channel
> Registration is the only stage in OPAQUE that requires a server-authenticated and confidential channel: either physical, out-of-band, PKI-based, etc.

I think the channel also has to protect integrity?

chris-wood:
Closes #393

@hugokraw -- can you please confirm this is correct?


hugokraw (commented):
Yes, integrity is needed. In an active  man-in-the-middle scenario there is no confidentiality without integrity (in the sense of confidentiality = authenticated encryption). If we use confidentiality (generically rather than with an explicitly defined mechanism) elsewhere, it needs to be understood as including integrity or as "authenticated encryption".

bytemare (approved):

kevinlewi (approved):

PR #404, opened 2023-03-21
Add x25519 AKE test vectors

### LINKED ISSUE #238: Curve25519 support
We have a use case where a HSM is used to store the server private key. Currently almost all HSMs use the PKCS#11 standard, which doesn't have support for ristretto255 + DH. It is my understanding that OPAQUE had official support for X25519 in the past but dropped it in favor of ristretto255.

Is it possible to elaborate on the reasons for that decision? Is it still in scope to re-introduce it and provide test vectors?

I believe that OPAQUE already does support using any algorithm as long as it does fulfill the stated security requirements, but official support would still be very favorable.

It should be noted that HSMs can use P256 instead, which is already officially supported by OPAQUE, but many, including us, would like to use Curve25519 when possible.

chris-wood:
This did introduce some editorial changes to accommodate the x25519 AKE, but I don't think they're unreasonable. The actual reference implementation needs to be cleaned up somewhat, but I think we can at least use this to test for interop for the time being. I'll work on cleaning up the code in a bit, so, starting as a draft now.

Closes #238

cc @daxpedda


stef:
lgtm. thank you for this effort!

draft-irtf-cfrg-opaque.md
```diff
@@ -1385,22 +1385,67 @@ following fields:
...
+- ScalarMult(k, B): Implemented as scalar multiplication as described in
+  {{NISTCurves}}.
+
+#### Curve25519 Group
+
+This section describes the implementation of the AKE group based on Curve25519,
+as defined in {{?Curve25519=RFC7748}}.
+
+- DeriveAuthKeyPair(seed): This function is implemented by returning the private
+  key based on seed (of length `Nseed = 32` bytes), as described in {{Section 5 of Curve25519}}.
```
bytemare, line 1443: This function also puts out the public key

draft-irtf-cfrg-opaque.md
```diff
@@ -1691,6 +1736,12 @@ See {{RFC5869}} for details.
 MAC of suitable length. For example, if MAC is HMAC-SHA256, then `Nh` could be
 32 bytes.
 
+## AKE Implementations
+
+This function is implemented as
+  DeriveKeyPair(seed, "OPAQUE-DeriveAuthKeyPair"), where DeriveKeyPair is
+  as specified in {{OPRF, Section 3.2}}.
```
bytemare, line 1743: I'm not sure why that phrase is here

poc/ake_group.sage
```diff
@@ -0,0 +1,276 @@
...
+    def __init__(self):
+        self.b_in_bytes = 64
+        self.bmax_in_bytes = 2^128
+        self.s_in_bytes = 136
+        self.name = "SHAKE-256"
+        
+    def hash(self,input_str, l = 64):
+        m = hashlib.shake_256(input_str)
+        digest = m.digest(l) # Note: hashlib.shake_256 seems to be buggy in some Sage environments :-(
+        return digest
```
bytemare, line 65: Do we need this?

bytemare (changes requested):

draft-irtf-cfrg-opaque.md
chris-wood: Leftover. Removed.

draft-irtf-cfrg-opaque.md
chris-wood: Yep, fixed!

daxpedda:
That's pretty amazing @chris-wood! :tada:

Gonna report back as soon as I figure out why I couldn't make them work in `opaque-ke`.

draft-irtf-cfrg-opaque.md
```diff
@@ -1385,22 +1385,68 @@ following fields:
...
+  {{NISTCurves}}.
+
+#### Curve25519 Group
+
+This section describes the implementation of the AKE group based on Curve25519,
+as defined in {{?Curve25519=RFC7748}}.
+
+- DeriveAuthKeyPair(seed): This function is implemented by returning the private
+  key `k` based on seed (of length `Nseed = 32` bytes), as described in {{Section 5 of Curve25519}},
+  as well as the result of ScalarMult(k, B), where B is the base point of Curve25519.
```
daxpedda, line 1444: Why does this differ from the other groups?
Because OPRF doesn't specify a ciphersuite for Curve25519?

chris-wood: Yeah, exactly.

daxpedda:
https://github.com/cfrg/draft-irtf-cfrg-opaque/pull/404/commits/0253cbb7f78a0b56fd183895556828da832aed74 updated the `draft-irtf-cfrg-opaque.md` but not the other files, I am assuming the new ones are correct?

poc/ake_group.sage
```diff
@@ -0,0 +1,49 @@
...
+        return scalar
+
+    def element_byte_length(self):
+        return 32
+
+    def scalar_byte_length(self):
+        return 32
+
+    def random_scalar(self, rng):
+        return os.urandom(32)
```
kevinlewi, line 35: @chris-wood: This needs to apply the clamping operations to the scalar, otherwise it can generate an invalid X25519 private key

cc: @daxpedda

chris-wood: Aha! Good catch.

chris-wood:
@bytemare can you please take another look?

draft-irtf-cfrg-opaque.md
daxpedda: After consulting the spec again, I couldn't really find an answer to this. Doesn't removing the `G.HashToScalar` have any security implications at all?
And if not, why is it in place for the other ciphersuites?

chris-wood: I'm not sure I understand the question. But to try and reiterate, there was no existing function in the OPRF document we could point to, which is why we just lean on standard key generation for curve25519 -- treat random bytes as the key (scalar).

daxpedda: Ah, sorry, I can see how my question is confusing. So in `opaque-ke` we were using OPRF with a custom ciphersuite when chossing Curve25519, which was possible because the hash2curve specification does provide a ciphersuite for Curve25519.

So my question is: the other ciphersuites do use OPRF to derive a key pair in `DeriveAuthKeyPair`. So why do other ciphersuites need that but Curve25519 doesn't? I thought it was necessary to for security reasons (that I have no clue about).

chris-wood: Oh, I see! You're asking if it's necessary to map the input seed to a scalar using something like a random oracle? I don't _think_ it's necessary, but I actually don't know the answer off the top of my head. I think this should probably be asked on the list (I can do that).

daxpedda: Thank you!

poc/rfc7748.sage
```diff
@@ -0,0 +1,209 @@
...
+    result = bytearray(num_bytes)
+    for i in range(num_bytes):
+    	result[i] = (Integer(u) >> 8*i) & 0xff
+    return bytes(result)
+
+def decodeScalar25519(k):
+    k_list = string_or_bytes_to_list(k)    
+    k_list[0] &= 248
+    k_list[31] &= 127
+    k_list[31] |= 64
```
kevinlewi, line 34: Sorry @chris-wood, I think actually the previous configuration was correct. In RFC 7748 it is done correctly. If we can copy these lines over to the scalar generation code, we should be good.

poc/ake_group.sage
kevinlewi: Let's revert to:

    key[0] &= 248
    key[31] &= 127
    key[31] |= 64
 
 as was originally suggested. Sorry for the confusion!

poc/opaque_core.sage
```diff
@@ -53,39 +53,45 @@ class OPAQUECore(object):
...
+        pk = self.config.group.scalar_mult(sk, self.config.group.generator())
         return self.config.group.serialize(pk)
 
     def derive_group_key_pair(self, seed):
         return DeriveKeyPair(MODE_OPRF, self.config.oprf_suite.identifier, seed, _as_bytes("OPAQUE-DeriveAuthKeyPair"))
 
-    def create_cleartext_credentials(self, server_public_key, client_public_key, server_identity, client_identity):
+    def derive_auth_key_pair(self, seed):
+        if self.config.group.name == "x25519":
+            return seed, self.config.group.scalar_mult(seed, self.config.group.generator())
```
kevinlewi, line 64: @chris-wood Additionally, I believe we need to make sure that `seed` is clamped here, before using it. Otherwise, the private key that we generate will not be clamped (since it just equal to the seed).

poc/vectors/formatted.txt
```diff
@@ -240,6 +240,242 @@ e60
 
 #### Configuration
 
+~~~
+OPRF: ristretto255-SHA512
+Hash: SHA512
+KSF: Identity
+KDF: HKDF-SHA512
+MAC: HMAC-SHA512
+Group: x25519
```
kevinlewi, line 249: This should say `Group: curve25519`

kevinlewi (changes requested):
Based on above comments

kevinlewi:
Spec changes:
- Renamed "ScalarMult(k, B)" to "DiffieHellman(k, B)" to more accurately reflect that we are doing a DH operation in the places where it is mentioned in the spec
- Renamed the group's display name from "x25519" to "curve25519" in the test vectors

Sage POC code changes:
- Fixed clamping issue
- Added test to ensure that private key outputs for curve25519 are always clamped

< Not ready to merge just yet, still fixing a bug that was introduced when trying to merge with main branch>

kevinlewi (approved):
This is approved by me in case it wasn't clear :P

PR #405, opened 2023-03-21
Return client pk in Recovery functions

### LINKED ISSUE #384: Return values of the Recover and RecoverCredentials functions
The _ClientFinish_ function (Section 6.2.3) needs the _client_public_key_ if the _client_identity_ is empty. However, the _client_public_key_ is calculated in the _Recover_ function (Section 4.1.3). 
Would it therefore make sense that both the _Recover_ function and the _RecoverCredentials_ function (Section 6.3.2.3) return the _client_public_key_ so that it can also be used for the _AuthClientFinalize_ function (_client_identity_ = _client_public_key_ if _client_identity_ is empty)?

bytemare:
Closes #384.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -660,13 +660,13 @@ Recover
 
 Input:
 - randomized_pwd, a randomized password.
-- server_public_key, the encoded server public key for the AKE protocol.
 - envelope, the client's Envelope structure.
 - server_identity, the optional encoded server identity.
 - client_identity, the optional encoded client identity.
 
 Output:
 - client_private_key, the encoded client private key for the AKE protocol.
+- client_public_key, the encoded client public key for the AKE protocol.
 - export_key, an additional client key.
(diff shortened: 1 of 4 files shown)


bytemare:
At this point I'm wondering whether we should not simply bubble up the `cleartext_creds` from `Recover()` up to `ClientFinish()` and feed it to `AuthClientFinalize()` to avoid input arguments comments like

> the optional encoded xyyz identity, which is set to xyz_public_key if not specified.

chris-wood (commented):
Should we also update the reference implementation to match?

chris-wood:
> At this point I'm wondering whether we should not simply bubble up the cleartext_creds from Recover() up to ClientFinish() and feed it to AuthClientFinalize() to avoid input arguments comments like

Maybe! Would you be willing to draft a change so we can see what this looks like?

bytemare:
@caw done :)

chris-wood (approved):

PR #409, opened 2023-03-27
Add Hugo's suggestion about Thresholding

bytemare:
In #397 , @hugokraw suggested

> I suggest the following edit  (marked in italics)
>
> The authors suggest implementing the OPRF phase as a Threshold OPRF
> {{TOPPSS}}, effectively forcing an attacker to act online or to control at
> least t key shares, where t is the threshold number of shares necessary to
> recombine the secret OPRF key*, and only then be able to* run an offline
> dictionary attack.

This PR adds this.

Moreover, he suggested

> You may or may not comment also in the following point:
If the OPRF servers are separate from the authentication server then
finding all n shares still does not help since you cannot run the dictionary
attack without the server's database.

@hugokraw, do you mean _n_ like in "all shares that have been created" or "a threshold number of shares"?

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1910,7 +1910,7 @@ implementation considerations.
   protocol is secure. The choice of such inputs is up to the application.
 - {{JKX18}} comments on a defense against offline
   dictionary attacks upon server compromise or honest-but-curious servers.
-  The authors suggest implementing the OPRF phase as a Threshold OPRF {{TOPPSS}},
+  The authors suggest implementing the OPRF phase as a threshold OPRF {{TOPPSS}},
   effectively forcing an attacker to act online or to control at least t key
   shares (among the total n), where t is the threshold number of shares necessary
   to recombine the secret OPRF key, and only then be able to run an offline dictionary
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1912,9 +1912,12 @@ implementation considerations.
   dictionary attacks upon server compromise or honest-but-curious servers.
   The authors suggest implementing the OPRF phase as a Threshold OPRF {{TOPPSS}},
   effectively forcing an attacker to act online or to control at least t key
-  shares, where t is the threshold number of shares necessary to recombine
-  the secret OPRF key, and only then be able to run an offline dictionary attack.
-  This implementation only affects the server and change nothing for the client.
+  shares (among the total n), where t is the threshold number of shares necessary
+  to recombine the secret OPRF key, and only then be able to run an offline dictionary
+  attack. This implementation only affects the server and change nothing for the client.
+  Furthermore, if the Threshold OPRF servers holding these keys are separate from
+  the authentication server, then recovering all n shares would still not suffice
+  to run an offline dictionnary attack without access to the client record database.
   However, this mechanism is out of scope for this document.
 
 The following list enumerates notable differences and refinements from the original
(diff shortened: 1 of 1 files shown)


hugokraw:
I meant *n* like in "all shares that have been created" (equivalently, all
servers through which the OPRF key is shared)

chris-wood (changes requested):

draft-irtf-cfrg-opaque.md
```diff
@@ -1908,14 +1908,17 @@ implementation considerations.
...
+- {{JKX18}} comments on a defense against offline
   dictionary attacks upon server compromise or honest-but-curious servers.
   The authors suggest implementing the OPRF phase as a Threshold OPRF {{TOPPSS}},
   effectively forcing an attacker to act online or to control at least t key
-  shares, where t is the threshold number of shares necessary to recombine
-  the secret OPRF key and run an offline dictionary attack. This implementation only affects the server and changes
-  nothing for the client. However, this mechanism is out of scope for this document.
+  shares (among the total n), where t is the threshold number of shares necessary
+  to recombine the secret OPRF key, and only then be able to run an offline dictionary
+  attack. This implementation only affects the server and change nothing for the client.
```
chris-wood, line 1917:
```suggestion
  attack. This implementation only affects the server and changes nothing for the client.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1908,14 +1908,17 @@ implementation considerations.
...
+  Furthermore, if the Threshold OPRF servers holding these keys are separate from
```
chris-wood, line 1918:
```suggestion
  Furthermore, if the threshold OPRF servers holding these keys are separate from
```

chris-wood (approved):

draft-irtf-cfrg-opaque.md
bytemare: hm, I'm referring to the "Threshold OPRF {{TOPPSS}}" some lines above, where I used an uppercase T. Should I add the citation on this line or lowercase the top one?

chris-wood: I think we should use lowercase threshold in both cases. It's just an adjective for the OPRF, not a proper noun.

draft-irtf-cfrg-opaque.md
```diff
@@ -1908,14 +1908,17 @@ implementation considerations.
   of the derivation of OPRF keys via a single PRF. As long as the derivation
   of different OPRF keys from a single OPRF has different PRF inputs, the
   protocol is secure. The choice of such inputs is up to the application.
-- {{JKX18}} comments on a strong defense against essentially the only
-  inherent remaining weakness of a "strong aPAKE", namely, offline
+- {{JKX18}} comments on a defense against offline
   dictionary attacks upon server compromise or honest-but-curious servers.
   The authors suggest implementing the OPRF phase as a Threshold OPRF {{TOPPSS}},
```
bytemare, line 1913:
```suggestion
  The authors suggest implementing the OPRF phase as a threshold OPRF {{TOPPSS}},
```

draft-irtf-cfrg-opaque.md
bytemare: I committed the change, but then re-checked [in the Paper](https://eprint.iacr.org/2017/363.pdf), where they say "which we define as Threshold OPRF (T-OPRF)". I believe we should use uppercase to stick to the language used by the paper we're referring to

chris-wood: I don't really agree. Again, there is not just one Threshold OPRF, there can be many, so I'm going to fix this up on main.

stef: the paper calls it 2HashTDH, later papers call it tdh-op.

bytemare: True, you're certainly right

PR #416, opened 2023-06-28
replace "password file" with "registration record"

nikgraf:
The term "password file" is not used anywhere else in the spec. Usually it's referred to as `record` and the type is `RegistrationRecord`. To avoid confusion I suggest to change it to `registration record` in this particular place.


kevinlewi (approved):
Good catch, thank you!

PR #419, opened 2023-09-07
Adding in various editorial edits from Hugo

kevinlewi:


draft-irtf-cfrg-opaque.md
```diff
@@ -991,6 +1005,8 @@ The server state `ServerState` may have the following fields:
 
 - server_ake_state: The `ServerAkeState` defined in {{protocol-3dh}}.
 
+Both of these states are ephemeral, and should be erased after the protocol completes.
```
chris-wood, line 1008:
```suggestion
Both of these states are ephemeral and should be erased after the protocol completes.
```

draft-irtf-cfrg-opaque.md
```diff
@@ -1814,9 +1830,11 @@ between different clients that might otherwise share the same password.
...
 both the client side and the server side. In particular, a malicious server can
 attempt to simulate honest responses to learn the client's password.
-Implementations and deployments of OPAQUE SHOULD consider additional checks to
-mitigate this type of attack: for instance, by ensuring that there is a
-server-authenticated channel over which OPAQUE registration and login are run.
+While this constitutes an exhaustive online attack, hence as expensive as an
+online guessing attack from the client side, it is completely avoided in
+typical cases where the channel between client and server is authenticated,
+e.g., via TLS. In addition to the above, such a channel provides privacy of user
+information, including identity and envelope values.
```
chris-wood, line 1837:
```suggestion
online guessing attack from the client side, it can be mitigated in when the channel
between client and server is authenticated, e.g., using server-authenticated TLS.
In such cases, these online attacks are limited to clients and the authenticated server
itself. Moreover, such a channel provides privacy of user information, including identity
and envelope values.
```

I don't think this statement is true. A server that authenticates over TLS can still carry out this attack. This suggestion helps clarify that authentication only changes who can carry out the attack, but does not remove it entirely.

chris-wood (changes requested):

draft-irtf-cfrg-opaque.md
```diff
@@ -412,7 +426,7 @@ This specification uses a KDF with the following API and parameters:
   into `L` bytes of output keying material.
 - Nx: The output size of the `Extract()` function in bytes.
 
-This specification also makes use of a collision-resistant Message Authentication Code
+This specification also makes use of a random-key robust Message Authentication Code
```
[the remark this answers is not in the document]
kevinlewi: I added the clause, ", defined in {{notable-design-differences}}," to help clarify it. It is a bit weird pointing to a section on notable design differences, but that is indeed where it is now defined.

We could pull it out to its own subsection within Security Considerations, i.e. 

```
## Random-Key Robustness

The random-key robustness property that we rely on for a MAC is as follows: given two random keys k1 and k2, it is infeasible to find a message m such that MAC(k1, m) = MAC(k2, m).
```

it would be a short subsection...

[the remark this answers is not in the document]
chris-wood: Could we just inline the definition here instead?

draft-irtf-cfrg-opaque.md
```diff
@@ -412,8 +426,8 @@ This specification uses a KDF with the following API and parameters:
   into `L` bytes of output keying material.
 - Nx: The output size of the `Extract()` function in bytes.
 
-This specification also makes use of a collision-resistant Message Authentication Code
-(MAC) with the following API and parameters:
+This specification also makes use of a random-key robust Message Authentication Code
+(MAC), defined in {{notable-design-differences}}, with the following API and parameters:
```
chris-wood, line 430:
```suggestion
(MAC). The random-key robustness property states that, given two random keys k1 and k2, it is infeasible to find a message m such that MAC(k1, m) = MAC(k2, m). The API and parameters for the random-key robust MAC is as follows:
```

chris-wood (approved):

kevinlewi:
Thanks for the suggestions. I added the definition of a random-key robust MAC in the sentence right after where it is first mentioned, and then later in the Notable Design Differences section, the reference to this definition is made.

draft-irtf-cfrg-opaque.md
```diff
@@ -412,8 +426,10 @@ This specification uses a KDF with the following API and parameters:
   into `L` bytes of output keying material.
 - Nx: The output size of the `Extract()` function in bytes.
 
-This specification also makes use of a collision-resistant Message Authentication Code
-(MAC) with the following API and parameters:
+This specification also makes use of a random-key robust Message Authentication Code
+(MAC). The random-key robustness property states that, given two random keys k1 and k2,
+it is infeasible to find a message m such that MAC(k1, m) = MAC(k2, m).
+The API and parameters for the random-key robust MAC is as follows:
```
chris-wood, line 432:
```suggestion
The API and parameters for the random-key robust MAC are as follows:
```

chris-wood (approved):

PR #420, opened 2023-09-07
Changing RFCXXXX string to OPAQUEv1-

kevinlewi:
Based on Scott Fluhrer's mailing list comments:

"At one point, the draft uses “RFCXXXX” as a protocol identifier in the preamble; is that expected to be replaced by the assigned RFC number?  If so, I would suggest you add instructions to the RFC editors to that effect."


chris-wood (commented):
Alternatively, we could just use `"OPAQUEv1"` or whatever as the string, and avoid the RFC number dependency altogether? That would let us generate test vectors right now.

kevinlewi:
Yeah that might be nicer. I see that VOPRF does "OPRFv1-" as the context string -- perhaps we can just follow suit?

chris-wood:
> Yeah that might be nicer. I see that VOPRF does "OPRFv1-" as the context string -- perhaps we can just follow suit?

Yep, that's my suggestion.

kevinlewi:
Updated the string from "RFCXXXX" to "OPAQUEv1-". Also re-ran test vector generation

chris-wood (approved):

PR #425, opened 2023-09-25
Fix Preamble signature

### LINKED ISSUE #418: Incorrect signature of function `Preamble`
The signature of function `Preamble` in Section 6.4.2.1 is not aligned with its description nor its function body. The function has a parameter `ke2` that's not used inside the function. Values `credential_response`, `server_nonce`, and `server_public_keyshare` are used in the function body, but are not passed in as parameters.

Currently the signature is `Preamble(client_identity, ke1, server_identity, ke2)`, but it probably should be `Preamble(client_identity, ke1, server_identity, credential_response, server_nonce, server_public_keyshare)`

chris-wood:
Closes #418

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -1550,7 +1550,8 @@ Input:
 Output:
 - preamble, the protocol transcript with identities and messages.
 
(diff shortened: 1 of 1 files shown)


kevinlewi (approved):
LGTM!

Issue #426, opened 2023-09-26
Inconsistency about OPRF seed reusage


daxpedda:
https://github.com/cfrg/draft-irtf-cfrg-opaque/blob/66834054528d1daf4f68d730fb95c15d625006d0/draft-irtf-cfrg-opaque.md?plain=1#L487-L489

https://github.com/cfrg/draft-irtf-cfrg-opaque/blob/66834054528d1daf4f68d730fb95c15d625006d0/draft-irtf-cfrg-opaque.md?plain=1#L792-L793

It says "SHOULD", so I guess it might not actually conflict, but maybe the first statement should link to the second statement.

kevinlewi:
The first statement is really just intended to refer to the flexibility that we give to an implementor to not use the same seed for all clients. For instance, they could choose to use a unique seed per client (but of course that seed needs to stay the same between registration and login). I added a sentence which should hopefully help to clarify this in #427

chris-wood:
Closing as completed yb #427.

PR #427, opened 2023-09-28
Clarifying the use of multiple client seeds

kevinlewi:
Addresses #426

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -485,8 +485,10 @@ protocol; see {{configurations}} for details.
 The server chooses a pair of keys (`server_private_key` and `server_public_key`)
 for the AKE, and chooses a seed (`oprf_seed`) of `Nh` bytes for the OPRF.
 The server can use `server_private_key` and `server_public_key` with multiple
-clients and can opt to use multiple seeds (so long as they are kept consistent for
-each client).
+clients. The server can also opt to use different seeds for each client,
+so long as they are maintained across the registration and online AKE stages, and
+kept consistent for each client (since an inconsistent mapping of clients to seeds
+could leak information as described in {{preventing-client-enumeration}}).
 
 ## Offline Registration
 


daxpedda:
If the OPRF seed has to be the same to prevent client enumeration, I think it should be noted here.

chris-wood (approved):
This LGTM. I'm also OK with folding in @daxpedda's additional suggestion. Please feel free to merge as you see fit.

kevinlewi:
Updated the sentence to say:

The server can also opt to use different seeds for each client,
so long as they are maintained across the registration and online AKE stages, and
kept consistent for each client (since an inconsistent mapping of clients to seeds
could leak information as described in {{preventing-client-enumeration}}).

Hope that works!

daxpedda:
> (since an inconsistent mapping of clients to seeds
> could leak information as described in {{preventing-client-enumeration}})

Wouldn't that require consistent mapping between pre- and post-registered clients?
E.g. unregistered client X requires to have the same seed as registered client X, otherwise an attacker could notice the difference and determine that client X was not registered but is registered now.

The only way I see this could be done is by deriving a seed from the client identifier/identity, which seems rather problematic to me.

daxpedda:
To clarify though: I think the change LGTM, I was just worried about how implementers could interpret this.

kevinlewi:
> > (since an inconsistent mapping of clients to seeds
> > could leak information as described in {{preventing-client-enumeration}})
> 
> Wouldn't that require consistent mapping between pre- and post-registered clients? E.g. unregistered client X requires to have the same seed as registered client X, otherwise an attacker could notice the difference and determine that client X was not registered but is registered now.
> 
> The only way I see this could be done is by deriving a seed from the client identifier/identity, which seems rather problematic to me.

Yes, this requires a consistent mapping between pre and post registered clients. The situation I am thinking of is the following. Let's say the server sets the policy:
- if the client identity starts with an A, then always use seed X for registration login
- if the client identity starts with a B, then always use seed Y for registration and login
- ... and so on

The purpose of the sentence "The server can also opt to use different seeds for each client" is to indicate that it is indeed OK to set this kind of policy. This is not exactly deriving a seed dependent on the client identity, but choosing a seed based on the client identity, which we don't view as problematic.

So I guess my main question is: is the language clear enough to convey that this is what we mean, or is it just causing unnecessary confusion (and maybe we should get rid of the suggestion altogether)?

daxpedda:
> So I guess my main question is: is the language clear enough to convey that this is what we mean, or is it just causing unnecessary confusion (and maybe we should get rid of the suggestion altogether)?

I believe it is clear enough.

PR #433, opened 2023-10-11
Minor clarification about sending additional information

kevinlewi:
Adding the following sentence under application considerations:

- Additional information: After completing the online AKE stage, the server
  may choose to send additional information, encrypted under `session_key`,
  to the client.


kevinlewi:
After some offline discussion, seems like we can omit making this change as it just adds more confusion.

PR #435, opened 2023-11-16
Incorporating server identity into OPRF computation as a recommendation

kevinlewi:


bytemare:
Should we add this in the reference implementations and vectors ?

bytemare (approved):

Issue #434, opened 2023-10-22
checking of password strength


stef:
the section `Password Salt and Storage Implications` says the following:

> Some applications may require learning the client's password for enforcing password rules. Doing so invalidates this important security property of OPAQUE and is NOT RECOMMENDED.

i agree that this is not a good thing, and also should not be recommended, but maybe caveat this, with "depending on threat model, for example internal corporate deployments are possible exceptions were enforcement of strong password policies is mandated by regulation"

> Applications should move such checks to the client.

"malicious" clients can easily ignore password rules, and the server will never know about this. This can be a problem when corporate rules mandate strong passwords, but users want to circumvent this for whatever reason (mostly lazyness?) - this is well known rule in application security that the server should never trust any input validation by the client.

> Note that limited checks at the server are possible to implement, e.g., detecting repeated passwords upon re-registrations or password change.

i wonder how this would work, can someone enlighten me?

stef:
ah, i think the the mac of the envelope will not change if nothing changes on re-registration. if the prf key is not changed. but 1/ reregistration itself is not well specified and 2/ i think how and where prf key comes from in such a case is neither, i do not think its a good idea to not change the prf key on reregistration.

kevinlewi:
Regarding the enforcement of strong password policies because it is mandated by regulation: I agree that this is to some extent in conflict with OPAQUE, because it fundamentally does not allow the server to perform these password checks server-side. In fact, I feel like if it is absolutely required for these server-side checks to be done on the password, then OPAQUE should not be used. Therefore, I feel like the wording should stay as-is, without the caveat.

(However, in practice, I feel like doing client-side checks should still suffice. Perhaps this is dependent on the regulation, but I feel like if a client is specifically trying to get around client-side checks to make a weaker password, then this is the client's fault, not the server's. But that's just my opinion of course :) )

Some other comments:
1. Note that limited checks at the server are possible to implement, e.g., detecting repeated passwords upon re-registrations or password change. This would work by essentially simulating the login mechanism using the old record with the server and checking if it succeeds. Note that it requires the client to be honest and report the success, which in my opinion is sufficient, but perhaps if there are regulations which mandate this to be foolproof, would not suffice.
[...]

PR #437, opened 2023-12-05
added proposal to caveat the server-side registration warning

stef:
draft proposal to settle #434

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -2252,14 +2252,9 @@ offline dictionary attack to recover the original password.
 
 Some applications may require learning the client's password for enforcing password
 rules. Doing so invalidates this important security property of OPAQUE and is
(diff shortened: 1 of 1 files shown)


kevinlewi:
I'm hoping to cut down on the wording and avoid referencing corporate deployments. How about the following suggested text, instead? 

> Some applications may require learning the client's password for enforcing password
> rules. Doing so invalidates this important security property of OPAQUE and is
> NOT RECOMMENDED, unless it is not possible for applications to move such checks
> to the client. Note that limited checks at the server are possible to implement, e.g.,
> detecting repeated passwords upon re-registrations or password change.

Note that the delta is that we are caveating the "NOT RECOMMENDED", with the phrasing "unless it is not possible for applications to move such checks to the client" without elaborating the different scenarios where that might not be possible. I hope this still captures what you are aiming for!

kevinlewi (changes requested):
Marking as request changes for the above comment ^

bytemare (approved):

PR #439, opened 2024-01-16
changed offline registration -> registration

stef:
reviewing the document at a whole for the rglc, i notice that registration is consistently referred to as "offline registration" - as the meme goes "You Keep Using That Word, I Do Not Think It Means What You Think It Means". most people understand offline as something that doesn't need a network connection. and with opaque `offline` in the context of `registration` has a much more nuanced and different meaning, which is confusing for the casual reader and implementer. and thus i think there is two options, the more expensive let's explicitly define "offline" in this context (and not mix it up with the other use through out this document in the context of attacks) or simply and cheaply delete the offline adjective from the context of registrations. i think this change eliminates one source of confusion and while doing so decreases entropy.


kevinlewi (approved):
Added some small adjustments to fix tests and catch one more spot where offline registration is mentioned. LGTM

PR #449, opened 2024-03-14
Clarifying wording about multiple seeds per client

kevinlewi:
Resolving a slight wording ambiguity that could be interpreted not the way we intend. Servers should consistently assign one  seed for a client, but across clients, different seeds could be used, so long as they are kept consistent.

Spawned from discussion offline with Hugo Krawczyk who requested this change.

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -485,10 +485,11 @@ protocol; see {{configurations}} for details.
 The server chooses a pair of keys (`server_private_key` and `server_public_key`)
 for the AKE, and chooses a seed (`oprf_seed`) of `Nh` bytes for the OPRF.
 The server can use `server_private_key` and `server_public_key` with multiple
-clients. The server can also opt to use different seeds for each client,
-so long as they are maintained across the registration and online AKE stages, and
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -485,10 +485,11 @@ protocol; see {{configurations}} for details.
 The server chooses a pair of keys (`server_private_key` and `server_public_key`)
 for the AKE, and chooses a seed (`oprf_seed`) of `Nh` bytes for the OPRF.
 The server can use `server_private_key` and `server_public_key` with multiple
-clients. The server can also opt to use different seeds for each client,
-so long as they are maintained across the registration and online AKE stages, and
-kept consistent for each client (since an inconsistent mapping of clients to seeds
-could leak information as described in {{preventing-client-enumeration}}).
+clients. The server can also opt to use different seeds for each client
```
bytemare, line 488: I find "use different seeds for each client" can be confusing.

What do you think about "use a different seed for each client (i.e. each client should..." ?

kevinlewi: Sounds good to me, thanks for the suggestion!

bytemare (approved):

PR #457, opened 2024-06-12
Final editorial nits

chris-wood:


kevinlewi:
Can you also remove the TODO line:

```
<!--
TODO(caw): As part of https://github.com/cfrg/draft-irtf-cfrg-opaque/issues/312, address
the failure case that occurs when Blind fails, noting that this is an exceptional case that
happens with negligible probability
-->
```

on line 1944 of the file? (not sure why I couldn't make this comment directly on the line, sorry about that)

kevinlewi (approved):

PR #458, opened 2024-06-28
Adds inputs necessary for reproducing the fake test vectors.

armfazh:
While working on updating the [opaque-ts](https://github.com/cloudflare/opaque-ts/) implementation I found that some extra fields are needed to reproduce the fake test vectors.

In particular, this allows the client to invoke the first part of login, and in the second part, after receiving a fake response from the server,  it should fail.

Support for draft v16: https://github.com/cloudflare/opaque-ts/pull/19


kevinlewi (commented):
Hi @armfazh, sorry for the delays in replying to this. However, I don't think these extra parameters are necessary to add in. The parameters that you added in here --`password`, `client_nonce`, `blind_login` -- these are all used to create a `KE1` message, right? However, these test vectors can be verified because they already contain `KE1` as input, and the resulting `KE2` output is all that is needed to verify that the test vectors match.

Perhaps you are going one step further by ensuring that the client fails correctly. However, this is essentially guaranteed by the protocol itself and out of scope for the test vectors. Perhaps the portion of your code which is doing this check can be separated from the check against the test vectors?

Let me know if this makes sense...

README.md
```diff
@@ -27,7 +27,7 @@ This requires that you have the necessary software installed.  See
 | [opaque](https://github.com/bytemare/opaque/)                                   | Go       | main    | N/A          |
 | [libopaque](https://github.com/stef/libopaque)                                  | C        | main    | N/A          |
 | [ecc](https://github.com/aldenml/ecc)                                           | C        | main    | N/A          |
-| [opaque-ts](https://github.com/cloudflare/opaque-ts)                            | TypeScript | v07   | [voprf-ts](https://github.com/cloudflare/voprf-ts) |
+| [opaque-ts](https://github.com/cloudflare/opaque-ts)                            | TypeScript | main  | [voprf-ts](https://github.com/cloudflare/voprf-ts) |
```
kevinlewi, line 30: Happy to merge this change in, though!

kevinlewi:
@armfazh, closing this but feel free to reopen if you think it needs to be addressed

Issue #469, opened 2024-11-13
CleartextCredentials serialization is unspecified


iaik-jheher:
CleartextCredentials is specified as:
```
struct {
  uint8 server_public_key[Npk];
  uint8 server_identity<1..2^16-1>;
  uint8 client_identity<1..2^16-1>;
} CleartextCredentials;
```

An instance of `CleartextCredentials` is then concatenated with the envelope nonce to generate the `auth_tag`:
```
auth_tag =
    MAC(auth_key, concat(envelope_nonce, cleartext_credentials))
```

However, the document does not specify anywhere how to serialize a `CleartextCredentials` structure to bytes. As opposed to every other structure in the document, which has fixed field lengths (and can simply be serialized by concatenating the fields), `CleartextCredentials` has two fields with dynamic lengths. It is not immediately obvious how it should be serialized.

I was able to make an educated guess (based on the length constraint) that `CleartextCredentials` should be serialized as:
```
concat(
    server_public_key,
    I2OSP(len(server_identity), 2), server_identity,
    I2OSP(len(client_identity), 2), client_identity)
```

Which also then matches the provided test vectors. Nevertheless, this should be specified in the document.

kevinlewi:
@iaik-jheher Thank you for pointing this out! This has been addressed in #471 , please take a look and let me know if this looks good.

kevinlewi:
Closed by #471

PR #471, opened 2024-11-19
Adding missing encodings for CleartextCredentials and CustomLabel

kevinlewi:
Addresses #469

### DIFF
--- a/draft-irtf-cfrg-opaque.md
+++ b/draft-irtf-cfrg-opaque.md
@@ -697,7 +697,14 @@ def Store(randomized_password, server_public_key,
     CreateCleartextCredentials(server_public_key, client_public_key,
                                server_identity, client_identity)
   auth_tag =
-    MAC(auth_key, concat(envelope_nonce, cleartext_credentials))
+    MAC(auth_key, concat(
+      envelope_nonce,
+      server_public_key,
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-opaque.md
```diff
@@ -1586,7 +1593,7 @@ Expand-Label(Secret, Label, Context, Length) =
     Expand(Secret, CustomLabel, Length)
 ~~~
 
-Where CustomLabel is specified as:
+Where CustomLabel is specified and encoded (following Section 3.4 of {{?RFC8446}}) as:
```
iaik-jheher, line 1596: A reference to RFC8446 might potentially be overkill here.

You might also specify `CustomLabel` as the concatenation of `I2OSP(Length, 2)`, `I2OSP(len("OPAQUE-" + Label), 2)`, `"OPAQUE-"`, `Label`, `I2OSP(len(Context))`, and `Context`. This would also match how `Preamble` is specified in 6.4.2.1.

kevinlewi: Thanks! I contemplated doing it this way, but ultimately decided that referencing the section in RFC8446 was a more concise way to describe the concatenation. Hope that this is still clear enough.

